PR #2, opened 2015-03-27
Adds behavior to be able pass object suggestions to Autosuggest.

jstejada:
- Opposite to only being able to pass suggestions as strings or numbers, or multi sections objects
- If no suggestionRenderer function is passed, it will attempt to render displayKey of suggestion object if suggestion is an object

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,17 +59,17 @@ function(input, callback) {
 `<suggestions>` can be have one of the following three formats:
 
 * **Single section with no title:** Array of strings, e.g.: `['Mentone', 'Mentone East']`
-* **Single section with no title:** Array of objects with `displayKey` for
+* **Single section with no title:** Array of objects with matching `displayKey` for
   rendering, e.g.:
 ```javascript
+<Autosuggest suggestions={getSuburbs}, displayKey={'name'} />
+
 [{
   name: "Mentone",
-  data: "Anything",
-  displayKey: "Mentone"
+  data: "Anything"
 }, {
   name: "Altona Meadows",
-  data: "More data",
-  displayKey: "Alt. Meadows"
+  data: "More data"
 }]
 ```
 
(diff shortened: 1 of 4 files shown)


jstejada:
@moroshko , what do you think about this?

moroshko:
@jstejada Sorry it took me so long to find time to read your Pull Request thoroughly.

First of all, thanks a lot for your efforts (and tests!).

As far as I can see, you are interested to be able to pass a list of objects (rather than list of strings) when calling the `callback()` function, i.e.: `callback(null, listOfSuggestionObjects)`
I wonder what is the underlying reason for that. The only thing that comes to my mind is the ability to pass additional data to `suggestionRenderer()`. Currently, only the suggestion string is passed to it, while you might want to have a suggestion object instead to be able to render additional info. Am I right, or I totally misunderstood your intent?

Regarding the implementation itself, I'm not a big fan of adding extra config parameters like `displayKey`. We already have `suggestionRenderer()` who is responsible for rendering, so I'd to have a single point of responsibility when dealing with rendering a suggestion. For example, if `suggestions` are:

    [{ name: 'Mill Park', postcode: '3082' }, { name: 'Mentone', postcode: '3192' }]

we could have:

    <Autosuggest suggestionRenderer={renderSuggestion} suggestions={...} />
    
    function renderSuggestion(suggestion) { // assuming that 'suggestion' is an object
      return suggestion.name;
    }

This way we eliminate the need for the extra `displayKey` config option.
Do you find any problem with this approach?

Also, I have work in progress in the [`es6` branch](https://github.com/moroshko/react-autosuggest/tree/es6) which I plan to release in the next couple of days. Apart from moving to es6 classes, I implemented something very similar to what you did, I think. Would you mind to check this branch and let me know what you think?

moroshko:
@jstejada The es6 branch was merged to master

jstejada:
@moroshko thanks for replying! sorry I haven't had time too look at this until now.

Yes, basically I want to be able to pass a list of objects as suggestions to `callback` in order to have additional data available for rendering inside `suggestionRenderer`.

Yes, displayKey is not really necessary unless you were passing suggestion objects without a `suggestionRenderer` function, but I don't really see a use case for it currently.

I checked the es6 branch, and that is exactly what I need.
Thanks!

PR #3, opened 2015-04-06
Added change listener onInputChange

hypereng:
Added support for adding a change listener that will fire when value of input field is changed. This makes it possible to use the input field's value from outside autosuggest component.

### DIFF
--- a/dist/Autosuggest.js
+++ b/dist/Autosuggest.js
@@ -28,6 +28,7 @@ var findDOMNode = React.findDOMNode;
 
 var lastSuggestionsInputValue = null,
     guid = 0;
+var nop = function nop() {};
 
 var Autosuggest = (function (_Component) {
   function Autosuggest(props) {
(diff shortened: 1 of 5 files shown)


moroshko:
@hypereng Thanks for the Pull Request!

First of all, I'm curious to know why do you need to know when the autosuggest input field value changes. Do you have some kind of dependency between the autosuggest field and other element on the page?

Regarding the implementation itself:
Wouldn't it be nicer to use the existing `inputAttributes` API rather than adding an additional configuration option (`onInputChange`)?

```javascript   
let inputAttributes = {
  onChange: function(...) {
    ...
  }
};
```

```xml
<Autosuggest inputAttributes={inputAttributes} suggestions={getLocations} />
```

At the moment, this is not possible because `Autosuggest` sets its own `onChange()` handler on the input field. But, we could easily fix that. I'm just trying to avoid adding extra config fields if not absolutely necessary.

What do you think?

hypereng:
First of all, thanks for looking at my pull request.

The reason I want to be able to listen to input value changes is that I want to implement a "save as you type" method that auto-saves any changes.

I agree with you that my change was not consistent with the component's API.
Now I have introduced a new prop ```inputEventAttributes``` that allow a parent component to add an event listener on the input field.

What do you think of this change?

moroshko:
@hypereng 
* What value do we get by introducing `inputEventAttributes` rather than using `inputAttributes`?
* Do you really need `onBlur` and `onKeyDown` to implement "save as you type"? I'm just trying to keep the API to minimal, and avoid adding new config options for the sake of adding them. If there is a real use case, I'll be happy to add them.

nthtran:
IMO a use should be able to pass `onChange`, `onBlur` and `onKeyDown` to `inputAttributes`. These are normal props for an `input` so if you overwrite them you should provide a way to access them.

moroshko:
`inputAttributes.onChange` added to v1.12.0. See [docs](https://www.npmjs.com/package/react-autosuggest#inputAttributesOption).

Issue #7, opened 2015-04-14
onSuggestionFocused


jstejada:
Hi @moroshko , I need to add event listeners for `mousenter` and `mouseleave` on my suggestions. I'm currently binding the event listener to my react component returned by `suggestionRenderer`, like so (simplified):

```javascript

function handleMouseEnter() {
  ...
}

suggestionRenderer: function(suggestion) {
  return (
    <div onMouseEnter={handleMouseEnter}>
    ...
    </div>
  );
}
```

However, the event listener is never being called, and I'm getting the following react warning:
`Warning: Any use of a keyed object should be wrapped in React.addons.createFragment(object) before being passed as a child.`

I suspect it has to do with the current mouseenter and mouseleave event listeners which are setting the state of `<Autosuggest>` and causing a rerender.

Do you have any thoughts on the exact cause of this, or any workarounds? Do you think you could support an option like `onSuggestionMouseEnter`, or maybe something like `onSuggestionFocused` to account for the key up and key down events as well? Sort of like how `onSuggestionSelected` works.

Thanks!!

jstejada:
nevermind, I figured it out.. and it doesn't have to do with react-autosuggest. Thanks!

jstejada:
@moroshko actually, I made it work, but I can't handle the case when a suggestion is focused using the up/down keys. Would you consider supporting an option like `onSuggestionFocused` like I mentioned in my first comment?

jstejada:
I just submitted pull request #8 for this feature, so I will close this issue.

PR #9, opened 2015-04-15
Support for onSuggestionFocused and onSuggestionUnfocused options

jstejada:
Hi @moroshko  , I'm just submitting this tiny pull request to add support for a onSuggestionFocused callback option as I mention in #7 

Let me know what you think

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,19 +11,19 @@
 * <a href="//moroshko.github.io/react-autosuggest" target="_blank">Basic example</a><br>
 * <a href="//moroshko.github.io/react-autosuggest#Custom renderer" target="_blank">Custom renderer</a><br>
 * <a href="//moroshko.github.io/react-autosuggest#Multiple sections" target="_blank">Multiple sections</a>
-* <a href="//moroshko.github.io/react-autosuggest#2 or more characters" target="_blank">2 or more characters</a>
+* <a href="//moroshko.github.io/react-autosuggest#Events playground" target="_blank">Events playground</a>
 
 ## Features
 
-* [x] [Accessible][wai-aria] (you can use both mouse and keyboard)
-* [x] In-memory caching (we retrive suggestions for a given input only once)
-* [x] Support for delayed requests (if request comes back after user types another letter, it will be ignored)
-* [x] You can either display a [plain list of suggestions][basic-example], or have [multiple sections][multiple-sections]
-* [x] Full control over [rendering a suggestion](#suggestionRendererOption) (you can display extra data, images, whatever you want)
-* [x] Full control over [styling](#styling) (we just provide the mechanics and classes for you)
-* [x] Full control over [when to show the suggestions](#showWhenOption) (e.g. when user types 2 or more characters)
-* [x] Ability to [pass props to the input field](#inputAttributesOption) (e.g. initial value, placeholder)
-* [x] Hooks: [onSuggestionSelected](#onSuggestionSelectedOption)
+* [Accessible][wai-aria] (you can use both mouse and keyboard)
+* In-memory caching (we retrive suggestions for a given input only once)
+* Support for delayed requests (if request comes back after user types another letter, it will be ignored)
+* You can either display a [plain list of suggestions][basic-example], or have [multiple sections][multiple-sections]
+* Full control over [rendering a suggestion](#suggestionRendererOption) (you can display extra data, images, whatever you want)
+* Full control over [styling](#styling) (we just provide the mechanics and classes for you)
+* Full control over [when to show the suggestions](#showWhenOption) (e.g. when user types 2 or more characters)
+* Ability to [pass props to the input field](#inputAttributesOption) (e.g. initial value, placeholder)
+* Various events for maximum flexibility: [onSuggestionSelected](#onSuggestionSelectedOption) (more coming soon...)
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@
 * <a href="//moroshko.github.io/react-autosuggest" target="_blank">Basic example</a><br>
 * <a href="//moroshko.github.io/react-autosuggest#Custom renderer" target="_blank">Custom renderer</a><br>
 * <a href="//moroshko.github.io/react-autosuggest#Multiple sections" target="_blank">Multiple sections</a>
+* <a href="//moroshko.github.io/react-autosuggest#2 or more characters" target="_blank">2 or more characters</a>
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ function getSuggestions(input, callback) {
 * [`suggestionRenderer`](#suggestionRendererOption)
 * [`suggestionValue`](#suggestionValueOption)
 * [`showWhen`](#showWhenOption)
+* [`onSuggestionSelected`](#onSuggestionSelectedOption)
 * [`inputAttributes`](#inputAttributesOption)
 
 <a name="suggestionsOption"></a>
@@ -95,7 +96,7 @@ function(suggestion, input) {
 }
 ```
 
-* `suggestion` - The [\<suggestion>](#suggestion) (string or object)
+* `suggestion` - The [suggestion](#suggestion) to render (string or object)
 * `input` - The value of the input field (e.g.: `'Men'`). If user interacts using the Up or Down keys at the moment, it will be the value of the input field **prior** to those interactions.
 
 For example:
@@ -157,6 +158,24 @@ function showWhen(input) {
              showWhen={showWhen} />
 ```
 
+<a name="onSuggestionSelectedOption"></a>
+##### onSuggestionSelected (optional)
+
+This function will be called when suggestion is selected via mouse click or Enter. It has one parameter which is the selected [suggestion](#suggestion) (string or object).
+
+For example:
+
+```javascript
+function onSuggestionSelected(suggestion) { // In this example 'suggestion' is a string
+  console.log('Suggestion selected: [' + suggestion + ']');
+}
+```
+
+```xml
+<Autosuggest suggestions={getSuggestions}
+             onSuggestionSelected={onSuggestionSelected} />
+```
+
 <a name="inputAttributesOption"></a>
 ##### inputAttributes (optional)
 
@@ -182,13 +201,62 @@ let inputAttributes = {
 
 The `<Autosuggest />` component comes with no styles. You can use the following classes to style it:
 
-* `react-autosuggest` - Component's wrapper. It includes both the input field and the suggestions list.
-* `react-autosuggest__suggestions` - Suggestions list wrapper
-* `react-autosuggest__suggestions-section` - Suggestions section wrapper (exists only when displaying multiple sections)
-* `react-autosuggest__suggestions-section-name` - Suggestions section name wrapper (exists only when displaying multiple sections and `sectionName` is specified)
-* `react-autosuggest__suggestion` - Single suggestion wrapper
+* `react-autosuggest`
+* `react-autosuggest__suggestions`
+* `react-autosuggest__suggestion`
+* `react-autosuggest__suggestions-section`
+* `react-autosuggest__suggestions-section-name`
+* `react-autosuggest__suggestions-section-suggestions`
 
-Example: [`examples/src/Autosuggest.less`](https://github.com/moroshko/react-autosuggest/blob/master/examples/src/Autosuggest.less)
+An example can be found in [`examples/src/Autosuggest.less`](https://github.com/moroshko/react-autosuggest/blob/master/examples/src/Autosuggest.less)
+
+The following diagrams explain the classes above.
+
+#### No sections
+
+    +---| react-autosuggest |-------------------------+
+    |                                                 |
+    |  <input>                                        |
+    |                                                 |
+    |  +--| react-autosuggest__suggestions |-------+  |
+    |  |                                           |  |
+    |  |  +--| react-autosuggest__suggestion |--+  |  |
+    |  |  |                                     |  |  |
+    |  |  +-------------------------------------+  |  |
+    |  |                                           |  |
+    |  +-------------------------------------------+  |
+    |                                                 |
+    +-------------------------------------------------+
+    
+
+#### Multiple sections
+
+    
+    +---| react-autosuggest |----------------------------------------------------+
+    |                                                                            |
+    |  <input>                                                                   |
+    |                                                                            |
+    |  +--| react-autosuggest__suggestions |----------------------------------+  |
+    |  |                                                                      |  |
+    |  |  +--| react-autosuggest__suggestions-section |--------------------+  |  |
+    |  |  |                                                                |  |  |
+    |  |  |  +--| react-autosuggest__suggestions-section-name |---------+  |  |  |
+    |  |  |  |                                                          |  |  |  |
+    |  |  |  +----------------------------------------------------------+  |  |  |
+    |  |  |                                                                |  |  |
+    |  |  |  +--| react-autosuggest__suggestions-section-suggestions |--+  |  |  |
+    |  |  |  |                                                          |  |  |  |
+    |  |  |  |  +--| react-autosuggest__suggestion |-----------------+  |  |  |  |
+    |  |  |  |  |                                                    |  |  |  |  |
+    |  |  |  |  +----------------------------------------------------+  |  |  |  |
+    |  |  |  |                                                          |  |  |  |
+    |  |  |  +----------------------------------------------------------+  |  |  |
+    |  |  |                                                                |  |  |
+    |  |  +----------------------------------------------------------------+  |  |
+    |  |                                                                      |  |
+    |  +----------------------------------------------------------------------+  |
+    |                                                                            |
+    +----------------------------------------------------------------------------+
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -195,7 +195,15 @@ function onSuggestionSelected(suggestion, event) {
 <a name="onSuggestionFocusedOption"></a>
 ##### onSuggestionFocused (optional)
 
-This function will be called when suggestion is focused via mouse hover or up/down keys. It has one parameter which is the focused [suggestion](#suggestion) (string or object).
+This function will be called when suggestion is focused via mouse hover or Up/Down keys.
+
+```js
+function(suggestion) {
+  ...
+}
+```
+
+* [suggestion](#suggestion) - The focused suggestion
 
 For example:
 
@@ -213,7 +221,15 @@ function onSuggestionFocused(suggestion) { // In this example 'suggestion' is a 
 <a name="onSuggestionUnfocusedOption"></a>
 ##### onSuggestionUnfocused (optional)
 
-This function will be called when suggestion is unfocused via mouse hover or up/down keys. It has one parameter which is the unfocused [suggestion](#suggestion) (string or object).
+This function will be called when suggestion is unfocused via mouse hover, Up/Down keys, ESC key, when clicking outside `<Autosuggest />` input, or when a suggestion is selected.
+
+```js
+function(suggestion) {
+  ...
+}
+```
+
+* [suggestion](#suggestion) - The unfocused suggestion
 
 For example:
 
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ Now, open `http://localhost:3000/examples/dist/index.html`
 
 [MIT](http://moroshko.mit-license.org)
 
-[status-image]: https://img.shields.io/codeship/41810250-aa07-0132-fbf4-4e62e8945e03.svg
+[status-image]: https://img.shields.io/codeship/41810250-aa07-0132-fbf4-4e62e8945e03/master.svg
 [status-url]: https://codeship.com/projects/67868
 [npm-image]: https://img.shields.io/npm/v/react-autosuggest.svg
 [npm-url]: https://npmjs.org/package/react-autosuggest
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ function getSuggestions(input, callback) {
 * [`showWhen`](#showWhenOption)
 * [`onSuggestionSelected`](#onSuggestionSelectedOption)
 * [`onSuggestionFocused`](#onSuggestionFocusedOption)
+* [`onSuggestionUnfocused`](#onSuggestionUnfocusedOption)
 * [`inputAttributes`](#inputAttributesOption)
 
 <a name="suggestionsOption"></a>
@@ -198,7 +199,7 @@ This function will be called when suggestion is focused via mouse hover or up/do
 
 For example:
 
-```javascript
+```js
 function onSuggestionFocused(suggestion) { // In this example 'suggestion' is a string
   console.log('Suggestion focused: [' + suggestion + ']');
 }
@@ -207,6 +208,24 @@ function onSuggestionFocused(suggestion) { // In this example 'suggestion' is a 
 ```xml
 <Autosuggest suggestions={getSuggestions}
              onSuggestionFocused={onSuggestionFocused} />
+```
+
+<a name="onSuggestionUnfocusedOption"></a>
+##### onSuggestionUnfocused (optional)
+
+This function will be called when suggestion is unfocused via mouse hover or up/down keys. It has one parameter which is the unfocused [suggestion](#suggestion) (string or object).
+
+For example:
+
+```js
+function onSuggestionUnfocused(suggestion) { // In this example 'suggestion' is a string
+  console.log('Suggestion unfocused: [' + suggestion + ']');
+}
+```
+
+```xml
+<Autosuggest suggestions={getSuggestions}
+             onSuggestionUnfocused={onSuggestionUnfocused} />
 ```
 
 <a name="inputAttributesOption"></a>
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # React Autosuggest
 
-[WAI-ARIA compliant](http://www.w3.org/TR/wai-aria-practices/#autocomplete) React autosuggest component.
+[WAI-ARIA compliant][wai-aria] React autosuggest component.
 
 ## Live Examples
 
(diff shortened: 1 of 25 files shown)


moroshko:
@jstejada The Pull Request looks great! Could you just explain what are you trying to achieve using `onSuggestionFocused`? What's your use case for this prop?

jstejada:
Thanks! So basically the suggestions are meant to be added to a collection by users. The use case is basically that I want to show a "preview" visualization of a suggestion on another view when the suggestion is focused, and actually add the suggestion when it is selected.

moroshko:
@jstejada What is your expected behaviour in the following two cases?

1. Imagine you have the last suggestion focused and its preview is visible. When you press the `Down` key, none of the suggestions are focused. What should happen to the preview?
2. Imagine you mouse hover over the first suggestion. Its preview is visible. Now you move the mouse away. What should happen to the preview?

Maybe we also need something like `onSuggestionUnfocused`?

BTW, will your project be publicly available? If yes, I'll be interested to see the "preview on focus" thing.

jstejada:
@moroshko 
1. In this case, if none of the suggestions are focused it should remove the preview. If the autosuggest iterates over the suggestions circularly, it should display the preview of the first suggestion.

2. When the mouse is moved away it should remove the preview.

How I am currently handling this is that every time I focus on a new suggestion I remove the currently displayed preview before displaying the new one. I'm also removing the preview when the mouse leaves a suggestion to somewhere different from another suggestion.

Definitely an option like `onSuggestionUnfocused` would be a much cleaner way of solving this, since my current solution is kind of hackish. I can go ahead and implement it in this pull request if you think its a good idea.

Absolutely, I can show you once its released, or maybe a screencast or something.

jstejada:
@moroshko I implemented what `onSuggestionUnfocused` would look like in another branch on my fork: https://github.com/jstejada/react-autosuggest/commits/on-suggestion-unfocused

Let me know what you think and I can merge into this PR

moroshko:
@jstejada 

If having `onSuggestionUnfocused` makes your life easier and the code is cleaner, than we should have it. I would like `react-autosuggest` to provide the best developer experience.

Please go ahead and merge your `on-suggestion-unfocused` branch to your master. I'll review then the updated Pull Request.

Will your work be publicly available? If yes, and you keen, I can put a link to your work in a "Used By" section of `react-autosuggest` (this section doesn't exist yet).

BTW, the reason pressing Down doesn't move from last suggestion to the first one is this ([WAI-ARIA guide](http://www.w3.org/TR/wai-aria-practices/#autocomplete)):
> Note that the arrows will wrap through the textbox when the top or bottom of the list is reached. For example, pressing the down arrow when the last choice is highlighted will move focus back to the textbox, pressing down again will move focus to the first item in the list. Likewise, with focus in the textbox and the list displayed, pressing up arrow will move focus to the last item in the list.

jstejada:
Great @moroshko , thanks!  I just merged in the updates from my `on-suggestion-unfocused` branch and also updated my fork with the latest upstream changes from this repo.

Yes, my app will be publicly available, so I'll let you know as soon as its released so you can add it to the "Used by" section

jstejada:
@moroshko , running the tests seems to be broken with the addition of experimental stage 0 for babel. I'm getting a syntax error when running `npm test`:

```js
  10 | export default class Autosuggest extends Component {
> 11 |   static propTypes = {
```

Are you able to run the tests?

src/Autosuggest.js
```diff
@@ -140,6 +145,10 @@ export default class Autosuggest extends Component {
     }
 
     this.setState(newState);
+
+    this.props.onSuggestionUnfocused(lastFocusedSuggestion);
```
moroshko: I believe `onSuggestionUnfocused` shouldn't be called if `lastFocusedSuggestion === null`. What do you think?

src/Autosuggest.js
```diff
@@ -140,6 +145,10 @@ export default class Autosuggest extends Component {
...
+    lastFocusedSuggestion = suggestion;
+    this.props.onSuggestionFocused(suggestion);
```
moroshko: I'd change the order of these 2 lines just to improve readability.

src/Autosuggest.js
```diff
@@ -211,10 +222,14 @@ export default class Autosuggest extends Component {
...
 
   onSuggestionMouseEnter(sectionIndex, suggestionIndex) {
+    let suggestion = this.getSuggestion(sectionIndex, suggestionIndex);
     this.setState({
       focusedSectionIndex: sectionIndex,
       focusedSuggestionIndex: suggestionIndex
     });
+
+    lastFocusedSuggestion = suggestion;
+    this.props.onSuggestionFocused(suggestion);
```
moroshko: Same, change order of these 2 lines?

src/Autosuggest.js
```diff
@@ -126,6 +130,7 @@ export default class Autosuggest extends Component {
 
   focusOnSuggestion(suggestionPosition) {
     let [sectionIndex, suggestionIndex] = suggestionPosition;
+    let suggestion = this.getSuggestion(sectionIndex, suggestionIndex);
```
moroshko: I wonder what will happen if `suggestionIndex === null`. This will happen when first suggestion is focused and Up is clicked, or the last suggestion is focused and Down is clicked. In this case we shouldn't be calling `onSuggestionFocused`, I think.

src/tests/Autosuggest.js
```diff
@@ -372,7 +382,7 @@ describe('Autosuggest', function() {
       it('should not call onSuggestionSelected when mouse enters a suggestion', function() {
         mouseOverFromInputToSuggestion(0);
         expect(onSuggestionSelected).not.toBeCalled();
-      });      
+      });
```
moroshko: BTW, Thanks for all these cleanups!

src/tests/Autosuggest.js
```diff
@@ -412,6 +422,93 @@ describe('Autosuggest', function() {
...
+          <Autosuggest suggestions={getSuburbStrings}
+                       onSuggestionFocused={onSuggestionFocused}
+                       onSuggestionUnfocused={onSuggestionUnfocused} />
+        );
+        setInputValue('m');
+      });
+
+      it('should call onSuggestionFocused when suggestion is focused using mouse', function() {
+        mouseOverFromInputToSuggestion(1);
+        expect(onSuggestionFocused).toBeCalledWith('Mordialloc');
```
moroshko: Make sure also that `onSuggestionUnfocused` wasn't called?

src/tests/Autosuggest.js
```diff
@@ -412,6 +422,93 @@ describe('Autosuggest', function() {
...
+      });
+
+      it('should call onSuggestionUnfocused when suggestion is unfocused using ESC key', function() {
+        clickDown();
+        clickEscape();
+        expect(onSuggestionUnfocused).toBeCalledWith({ suburb: 'Mill Park', postcode: '3083' });
+      });
+    });
+  });
+
```
moroshko: I don't like how we end up duplicating tests in "String suggestions" and "Object suggestions".
Also, I'd prefer to split the tests for `onSuggestionFocused` and `onSuggestionUnfocused`. This way, if you have an issue with, say `onSuggestionUnfocused`, you don't have to read the `onSuggestionFocused` stuff (amount of code in tests that you have to check is optimal). How about we have a section for `onSuggestionFocused` tests which will deal with "string suggestions" and another section for `onSuggestionUnfocused` tests with "object suggestions"? (If you end up having a bug with "object suggestions" in `onSuggestionFocused`, for example, we might not catch it, but I'd say that the chance that this will happen is very low, and I'll be happy to take the risk to make our tests DRY).

Also, we should have tests for edge cases: first suggestion is focused and Up is clicked, and last suggestion is focused and Down clicked.

One thing that we potentially missed here is if user clicks away from the input field (`onInputBlur` is called in this case). We should check if suggestion was focused, and if yes call `onSuggestionUnfocused`.

moroshko:
@jstejada I just cloned your latest master and tests are running perfectly fine. Maybe try to restart your shell?

jstejada:
@moroshko great! I will look over your code review and probably push the changes today.

I'll  see whats going on with the tests, maybe it has something to do with my babel install.

Thanks!

src/Autosuggest.js
jstejada: yes, that probably makes better sense

jstejada:
hi @moroshko , so several points:

- Managed to get the tests running again
- Updated ordering of code for better clarity and readability
- Now calling `onSuggestionUnfocused` when clicking outside, i.e. on input `blur`
- Accounted for edge cases you mentioned, as well as a couple of others (should not call `onSuggestionUnfocused` when pressing ESC and no suggestion currently focused, same goes for on input `blur`)
- Updated test structure as you proposed, and added tests for edge cases.

Let me know!

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,106 @@ describe('Autosuggest', function() {
...
   });
 
+  describe('onSuggestionFocused', function() {
+    beforeEach(function() {
+      onSuggestionFocused.mockClear();
+      onSuggestionUnfocused.mockClear();
+      createAutosuggest(
+        <Autosuggest suggestions={getSuburbStrings}
+                     onSuggestionFocused={onSuggestionFocused}
+                     onSuggestionUnfocused={onSuggestionUnfocused} />
```
moroshko: I suggest that in the `onSuggestionFocused` section we don't care about `onSuggestionUnfocused` at all.

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,106 @@ describe('Autosuggest', function() {
...
+      );
+      setInputValue('m');
+    });
+
+    describe('Mouse interactions', function() {
+
+      it('should call onSuggestionFocused when suggestion focused using mouse', function() {
+        mouseOverFromInputToSuggestion(1);
+        expect(onSuggestionFocused).toBeCalledWith('Mordialloc');
+        expect(onSuggestionUnfocused).not.toBeCalled();
```
moroshko: We shouldn't be testing for `onSuggestionUnfocused` in the `onSuggestionFocused` section. This is the whole point of having 2 separate sections.

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,106 @@ describe('Autosuggest', function() {
...
+
+  describe('onSuggestionUnfocused', function() {
+    beforeEach(function() {
+      onSuggestionFocused.mockClear();
+      onSuggestionUnfocused.mockClear();
+      createAutosuggest(
+        <Autosuggest suggestions={getSuburbObjects}
+                     suggestionRenderer={renderSuburbObject}
+                     suggestionValue={getSuburbObjectValue}
+                     onSuggestionFocused={onSuggestionFocused}
```
moroshko: Same. Let's not set `onSuggestionFocused` here (and also remove `onSuggestionFocused.mockClear();` above).

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,106 @@ describe('Autosuggest', function() {
...
+      it('should call onSuggestionUnfocused when clicking outside and suggestion is focused', function() {
+        mouseOverFromInputToSuggestion(0);
+        clickOutside();
+        expect(onSuggestionUnfocused).toBeCalledWith({ suburb: 'Mill Park', postcode: '3083' });
+      });
+
+      it('should not call onSuggestionUnfocused when clicking outside and no suggestion focused', function() {
+        clickOutside();
+        expect(onSuggestionUnfocused).not.toBeCalled();
+      });
```
moroshko: I suggest we move the above 2 tests to the `'Keyboard interactions'` section below because you can't really click outside without moving the mouse from the suggestion to the outside area first. It makes more sense to me to check a scenario when you:
1. click Down to focus on suggestion
2. click outside
3. verify onSuggestionUnfocused is called
or 
1. click Down
2. click Down
3. click Down (now focus should go back to the input field and no suggestion is focused)
4. click outside
5. verify onSuggestionUnfocused is not called

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,106 @@ describe('Autosuggest', function() {
...
+        clickEscape();
+        expect(onSuggestionUnfocused).toBeCalledWith({ suburb: 'Mill Park', postcode: '3083' });
+      });
+
+      it('should not call onSuggestionUnfocused when ESC key pressed and no suggestion focused', function() {
+        clickEscape();
+        expect(onSuggestionUnfocused).not.toBeCalled();
+      });
+    });
+  });
```
moroshko: I wonder what do you expect to happen when suggestion is selected via mouse click or via clicking Enter? Shall we call `onSuggestionUnfocused` in both cases? Could you please add tests for those?

src/Autosuggest.js
```diff
@@ -140,6 +145,18 @@ export default class Autosuggest extends Component {
     }
 
     this.setState(newState);
+
+    this.onSuggestionUnfocused();
+    if(suggestion != null) {
```
moroshko: I prefer we use strict equality to keep things as explicit as possible. In this scenario, I believe `suggestion` can be `undefined` because that's what `getSuggestion(sectionIndex, suggestionIndex)` will return when `suggestionIndex === null`. The convention that I've been using to identify whether a suggestion is focused is to check whether `suggestionIndex === null`, so let's keep with that. We could have a `isSuggestionFocused()` function, if that make sense to add it.

moroshko:
@jstejada I finished another round of code review. Looks great, though there are still few little things to address. Thanks a lot for your time. You are such a great contributor! Pleasure to work with you :)

src/tests/Autosuggest.js
jstejada: yep, missed that

src/tests/Autosuggest.js
jstejada: missed it here too

src/tests/Autosuggest.js
jstejada: sounds good

src/tests/Autosuggest.js
jstejada: yes, I forgot about this one

src/Autosuggest.js
jstejada: cool, will update this

jstejada:
thanks @moroshko !! yes, a few details are missing. I will go over them soon.

jstejada:
@moroshko i just pushed the changes based on your latest code review. Basically I:

- Cleaned up the tests according to your comments
- Added behavior to call `onSuggestionUnfocused` when selecting a suggestion + tests

README.md
```diff
@@ -56,6 +56,8 @@ function getSuggestions(input, callback) {
 * [`suggestionValue`](#suggestionValueOption)
 * [`showWhen`](#showWhenOption)
 * [`onSuggestionSelected`](#onSuggestionSelectedOption)
+* [`onSuggestionFocused`](#onSuggestionFocusedOption)
+* [`onSuggestionUnfocused`](#onSuggestionUnfocusedOption)
```
moroshko: We should also add these 2 to the "Features" section above.

src/Autosuggest.js
```diff
@@ -208,13 +237,19 @@ export default class Autosuggest extends Component {
 
   onInputBlur() {
     this.setSuggestionsState(null);
+    this.onSuggestionUnfocused();
+    lastFocusedSuggestion = null;
```
moroshko: Why do we need to set `lastFocusedSuggestion` to `null` here? I tried to remove this line and all the tests still pass. Do we have a test that is missing, or this line is redundant?

src/Autosuggest.js
```diff
@@ -183,6 +210,8 @@ export default class Autosuggest extends Component {
         }
 
         this.setState(newState);
+        this.onSuggestionUnfocused();
+        lastFocusedSuggestion = null;
```
moroshko: I feel like we shouldn't be setting `lastFocusedSuggestion` outside of `this.onSuggestionUnfocused()` and `this.onSuggestionFocused()` (this one doesn't exist at the moment). What do you think?

src/Autosuggest.js
```diff
@@ -208,13 +237,19 @@ export default class Autosuggest extends Component {
...
 
   onSuggestionMouseEnter(sectionIndex, suggestionIndex) {
+    let suggestion = this.getSuggestion(sectionIndex, suggestionIndex);
     this.setState({
       focusedSectionIndex: sectionIndex,
       focusedSuggestionIndex: suggestionIndex
     });
+
+    this.props.onSuggestionFocused(suggestion);
+    lastFocusedSuggestion = suggestion;
```
moroshko: Yeah, I feel like we better have `this.onSuggestionFocused()` (similarly to `this.onSuggestionUnfocused()`) and set `lastFocusedSuggestion` there.

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,117 @@ describe('Autosuggest', function() {
...
+      onSuggestionFocused.mockClear();
+      createAutosuggest(
+        <Autosuggest suggestions={getSuburbStrings}
+                     onSuggestionFocused={onSuggestionFocused} />
+      );
+      setInputValue('m');
+    });
+
+    describe('Mouse interactions', function() {
+
```
moroshko: Remove empty line?

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,117 @@ describe('Autosuggest', function() {
...
+      it('should call onSuggestionFocused when suggestion focused using mouse', function() {
+        mouseOverFromInputToSuggestion(1);
+        expect(onSuggestionFocused).toBeCalledWith('Mordialloc');
+      });
+    });
+
+    describe('Keyboard interactions', function() {
+
```
moroshko: Remove empty line?

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,117 @@ describe('Autosuggest', function() {
...
+        mouseOverFromInputToSuggestion(0);
+        mouseOverBetweenSuggestions(0, 1);
+        expect(onSuggestionUnfocused).toBeCalledWith({ suburb: 'Mill Park', postcode: '3083' });
+      });
+
+      it('should call onSuggestionUnfocused when suggestion selected with mouse', function() {
+        mouseOverFromInputToSuggestion(0);
+        mouseDownSuggestion(0);
+        expect(onSuggestionUnfocused).toBeCalledWith({ suburb: 'Mill Park', postcode: '3083' });
+      });
```
moroshko: I think we are missing 2 tests here:
1. Suggestions are shown, one of them is focused, and user clicks outside. `onSuggestionUnfocused` should be called.
2. Suggestions are shown, none of them are focused, and user clicks outside. `onSuggestionUnfocused` should not be called.

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,117 @@ describe('Autosuggest', function() {
...
+    });
+
+    describe('Keyboard interactions', function() {
+
+      it('should call onSuggestionUnfocused when suggestion unfocused using Up/Down keys', function() {
+        clickDown();
+        clickDown();
+        expect(onSuggestionUnfocused).toBeCalledWith({ suburb: 'Mill Park', postcode: '3083' });
+      });
```
moroshko: We also need a test for this scenario:
1. user clicks ESC (suggestions disappear)
2. user clicks Up/Down (suggestions appear, but none are focused)
`onSuggestionUnfocused` should not be called.

moroshko:
@jstejada Have a look at the [Events playground](http://moroshko.github.io/react-autosuggest/#Events playground) example, which currently have `onSuggestionSelected` only.
It would be great if you could add `onSuggestionFocused` and `onSuggestionUnfocused` to this example. It would be much easier to play with it and find bugs/missing tests ;)

src/tests/Autosuggest.js
jstejada: you mean `onSuggestionFocused`?

moroshko: No. This section is for `onSuggestionUnfocused`. The test above checks that `onSuggestionUnfocused` is called after Up/Down, which is not true in case ESC was pressed before.

jstejada: oh ok i got it, I'll add it shortly

jstejada: it should also apply for `onSuggestionFocused` right? When the user clicks Up/Down  it shouldn't focus any suggestion

jstejada:
@moroshko So I added a few things based on your comments:

- Cleaned up redundant code and empty lines, and added a `Autosuggest#onSuggestionFocused` to wrap that logic.
- Added the test for the case of pressing Up/Down keys after ESC.
- Updated the README's Features section
- Updated the Events Playground example with `onSuggestionFocused` and `onSuggestionUnfocused`
  - For this one I had to update `Example#onEventAdded` because when events got added too quickly (which is the case when moving through suggestions, i.e. focusing and unfocusing), the component state didn't get updated properly since `setState` is asynchronous. So basically I added an event queue to serialize adding events to state.

Let me know what you think! Thanks!

examples/src/Examples.js
```diff
@@ -11,6 +11,8 @@ import MultipleSections from './MultipleSections/MultipleSections';
 import EventsPlayground from './EventsPlayground/EventsPlayground';
 import EventsLog from './EventsLog/EventsLog';
 
+let eventQueue = [];
```
moroshko: I'd prefer to have `eventQueue` in the class itself (the same way we have `eventsPlaceholder`).

examples/src/Examples.js
```diff
@@ -83,10 +85,23 @@ export default class Examples extends Component {
     return null;
   }
 
+  processEvents() {
+    if (eventQueue.length > 0) {
+      let event = eventQueue[0];
```
moroshko: Could do `let event = eventQueue.shift()` here, and in the callback below get rid of `eventQueue.shift();`.

src/Autosuggest.js
```diff
@@ -140,6 +144,33 @@ export default class Autosuggest extends Component {
...
+    lastFocusedSuggestion = null;
+  }
+
+  onSuggestionSelected(sectionIndex, suggestionIndex, event) {
+    this.onSuggestionUnfocused();
+    this.props.onSuggestionSelected(
+      this.getSuggestion(sectionIndex, suggestionIndex),
+      event
+    );
+    lastFocusedSuggestion = null;
```
moroshko: Why this is required? Wouldn't `lastFocusedSuggestion` be set to `null` when `this.onSuggestionUnfocused()` is called?

src/Autosuggest.js
```diff
@@ -215,6 +249,8 @@ export default class Autosuggest extends Component {
       focusedSectionIndex: sectionIndex,
       focusedSuggestionIndex: suggestionIndex
     });
+
+    this.onSuggestionFocused(sectionIndex, suggestionIndex);
```
moroshko: Shouldn't we call `this.onSuggestionUnfocused()` instead here? If so, how come tests didn't catch this?

src/tests/Autosuggest.js
moroshko: That's right. But, we already have the equivalent test for `onSuggestionFocused`.

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,129 @@ describe('Autosuggest', function() {
...
+        mouseDownSuggestion(0);
+        expect(onSuggestionUnfocused).toBeCalledWith({ suburb: 'Mill Park', postcode: '3083' });
+      });
+    });
+
+    describe('Keyboard interactions', function() {
+      it('should call onSuggestionUnfocused when suggestion unfocused using Up/Down keys', function() {
+        clickDown();
+        clickDown();
+        expect(onSuggestionUnfocused).toBeCalledWith({ suburb: 'Mill Park', postcode: '3083' });
```
moroshko: I'd also be inclined to check that `onSuggestionUnfocused` isn't called after the first `clickDown();`. If that was the case, this test would still pass.

src/tests/Autosuggest.js
```diff
@@ -412,6 +426,129 @@ describe('Autosuggest', function() {
...
+      it('should not call onSuggestionUnfocused when clicking outside and no suggestion focused', function() {
+        clickDown();
+        clickDown();
+        clickDown();
+        onSuggestionUnfocused.mockClear();
+        clickOutside();
+        expect(onSuggestionUnfocused).not.toBeCalled();
+      });
+
+      it('should call onSuggestionUnfocused when suggestion selected with Enter key', function() {
```
moroshko: How about renaming it to:

    should call onSuggestionUnfocused when Enter is pressed and suggestion is focused

and add a test for:

    should not call onSuggestionUnfocused when Enter is pressed and no suggestion focused

moroshko:
@jstejada I played with the updated Events playground (thanks!), and I believe I found a bug:

1. Press 'm'
2. Click down (first suggestion is focused)
3. Mouse over this same first suggestion from outside

I believe that `onSuggestionFocused` shouldn't be called in this case. Would you agree?

Could you please a test for this scenario?

moroshko:
@jstejada Here is another bug:

1. Press 'a'
2. Press down (first suggestion is focused)
3. Press Backspace

There is no focused suggestion now, but `onSuggestionUnfocused` hasn't been called.

We should probably call `onSuggestionUnfocused` at the end of `onInputChange`.

Could you please verify that every time `focusedSuggestionIndex` is set, we either call `onSuggestionUnfocused` (when it is set to `null`) or we call `onSuggestionFocused` (when it is set to a number >= 0).

It would be nice to have test for this scenario as well.

moroshko:
@jstejada Finished another round of code review :) We are getting there!

examples/src/Examples.js
jstejada: the thing is I only want to remove it from the queue when it is actually done processing it

src/Autosuggest.js
jstejada: I don't think so. When the mouse enters we should call `onSuggestionFocused`, and when the mouse leaves we should call `onSuggestionUnfocused`. That is why the tests are working. We could implement all of the logic inside this handler though.

moroshko: Hmm.. I'm sure I saw this being called in `onSuggestionMouseLeave()`. Must be a glitch in GitHub or my head :) Please ignore.

jstejada:
@moroshko great! I just pushed some changes. Basically:

- Removed that redundant line and updated the tests.
- Handled the 2 cases you discovered using the events playground

examples/src/Examples.js
```diff
@@ -83,10 +86,23 @@ export default class Examples extends Component {
...
+      });
+    }
+  }
+
   onEventAdded(event) {
-    this.setState({
-      events: this.eventsExist() ? this.state.events.concat([event]) : [event]
-    });
+    this.eventQueue.push(event);
+    if(this.eventQueue.length === 1) {
```
moroshko: @jstejada This whole `processEvents()` thing is still a bit of a mystery to me. Why are we processing events only if there is a single item in the queue? I understand that we may get 2 events at the same time (`unfocus + focus`, or `unfocus + select`), but why to process after the first one? Could you please elaborate on this?

jstejada: Absolutely. Since every event is added to the queue (on line 102), that means that if the queue length is 1, then the queue was empty. It should **only** call `processEvents` when the queue is empty because if the queue is not empty it means that a `processEvents` loop is already running on the queue (it will run while there are events in the queue left to process) and I don't want to start another one, just add the new event to the queue.

The initial problem was that since unfocusing and focusing happens so fast while moving between suggestions, it ended up only recording the focus event and not the unfocus event. I figured that if you call `setState` before another call to `setState` has completed and it is modifying the same state, only the last call to `setState` will apply, because that is what I was observing. In any case, I implemented the queue to serialize the event adding process and only add new events after all of the others have been successfully added.

I tried to implement `onEventAdded` as least verbose as possible, but I think it affects readability and can be confusing, so I can reimplement to explicitly check for an empty queue.

moroshko:
@jstejada Thanks for your contribution and patience :+1:

jstejada:
@moroshko great! thank you for considering my pull request and taking the time to review it!

moroshko:
It's on npm v1.11.0

jstejada:
@moroshko awesome, thanks!! :smile:

Issue #10, opened 2015-04-15
Always call event.preventDefault on onInputKeyDown?


stream7:
Hello!

I have a use case where when enter is clicked in order to select one suggestion the form is submitted. I think that it might be a good idea to expose the event onSuggestionSelected

```
function onSuggestionSelected(suggestion, event) {
  console.log('Suggestion selected: [' + suggestion + ']');
}
```

See https://github.com/moroshko/react-autosuggest/blob/master/src/Autosuggest.js#L181

moroshko:
@stream7 Thanks for #11! I think it's probably not a good idea to **always** call `preventDefault()` as users might want to submit a form when suggestion is selected. User should be in control whether to submit a form or not, and this is why you want the `event` object, right? You basically want to call `event.preventDefault()` inside `onSuggestionSelected()`, right?

stream7:
>  You basically want to call event.preventDefault() inside onSuggestionSelected(), right?

@moroshko yes that is correct

stream7:
thanks!

moroshko:
Thank you for the contribution! Hopefully, I'll publish a new version on npm in the next day or two.

moroshko:
It's in v1.10.0 now, enjoy!

stream7:
@moroshko Thank you! :)

w35l3y:
So should it be possible to call `event.preventDefault()` after this merge ?

I just tried to call it inside `onSuggestionSelected` and still the option was selected.

I am using version 9.3.2

moroshko:
@w35l3y Could you create a new issue with a Codepen that demonstrates the issue?

PR #11, opened 2015-04-15
Add event to onSuggestionSelected signature

stream7:
See https://github.com/moroshko/react-autosuggest/issues/10


moroshko:
closes #10

PR #15, opened 2015-04-29
Allow external access to input onchange events

icodejs:
Currently the AutoSuggest component wraps an input element, but
doesn't allow access to the onChange event.

This event is useful in situations where the user types some text that isn't
one of the suggested values but still needs to be captured by parent
component.


moroshko:
Duplicate of #3

moroshko:
@icodejs `inputAttributes.onClose` added to v1.12.0. See [docs](https://www.npmjs.com/package/react-autosuggest#inputAttributesOption).

icodejs:
Thanks @moroshko. Much appreciated!

PR #23, opened 2015-05-15
Allow to set a custom guid to the component

Ekhoo:
In a normal case, your component works fine, but in an isomorphic app, react display a warning because the server and client guid are different. You set a different guid (by incrementing it) for each component creation. In an isomorphic app, we need to have all of ours components equals. Without it, we are loosing the react benefit and the page will be re render totally. So we need to override it when it's necessary.

### DIFF
--- a/dist/Autosuggest.js
+++ b/dist/Autosuggest.js
@@ -46,8 +46,13 @@ var Autosuggest = (function (_Component) {
 
     _get(Object.getPrototypeOf(Autosuggest.prototype), 'constructor', this).call(this);
 
-    guid += 1;
-    this.id = guid;
+    if (props.inputAttributes.id) {
+       this.id = props.inputAttributes.id;
+     } else {
+       guid += 1;
+       this.id = guid;
+     }
+    
     this.cache = {};
     this.state = {
       value: props.inputAttributes.value || '',
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -30,8 +30,13 @@ export default class Autosuggest extends Component {
   constructor(props) {
     super();
 
-    guid += 1;
-    this.id = guid;
+    if (props.inputAttributes.id) {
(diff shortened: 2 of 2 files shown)


moroshko:
@Ekhoo Thanks for the Pull Request! I understand what you say, but can't clearly see why the generated guids are different on the server and the client.

Could you elaborate a bit more on the warning please? Which HTML element the warning refers to?

moroshko:
@Ekhoo I wonder what happened with the issue you experienced? Did you solve the issue?

Ekhoo:
Yes the issue is solved, but I removed some useless changes from the `dist/Autosuggest.js`

Ekhoo:
Is it ok for you ? Can you merge and make a new release on npm ?

Ekhoo:
If you try it yourself, you will see that `this.id` is different on the client and the server. I just add the ability to set a custom and static id.

moroshko:
@Ekhoo Do you have a working example that I can run to recreate the issue? 

Isn't `let guid = 0;` run both on the client and the server? What are the different values of `this.id` that you see on the client and the server?

Will be happy to merge the PR once I understand what's going on.

Ekhoo:
Ok I will explain. On the first render, client and server have the same GUID (1). Then, if I refresh the page, the GUID of the client will still be equals 1, whereas, the GUI of the server will be equals to 2 and so on.

moroshko:
@Ekhoo This makes perfect sense! 

Out of curiosity, what is the error that React throws?

moroshko:
@Ekhoo This should be solved in v1.13.0. There is no incremental guid anymore. `<Autosuggest />`'s id is static now (it defaults to `'1'`). [You should set the `id` if there are multiple `<Autosuggest />`s on a page.](https://github.com/moroshko/react-autosuggest#idOption)

Issue #24, opened 2015-05-18
Package dist as single file for client-side inclusion


liamzebedee:
I'm using Meteor which doesn't allow importing NPM packages on the client-side - is it possible to get this packaged into a single file for client-side inclusion.

I'm trying to hack a solution up in Webpack/Browserify now...

Thanks :)

moroshko:
@liamzebedee I prefer keep using Webpack rather than introducing another tool (Browserify). Could you check please if [this](https://github.com/moroshko/react-autosuggest/blob/standalone/dist/standalone/autosuggest.min.js) works?

liamzebedee:
Okay, makes sense and cheers!

I've loaded it, throw `Uncaught TypeError: Cannot read property 'Component' of undefined` where the snippet is something like `onSuggestionFocused:function(){},onSuggestionUnfocused:function(){},inputAttributes:{}},enumerable:!0}]),t}(g.Component);` 

Looks like the React variable is being minified to `g` and thus it can't access React.Component

moroshko:
@liamzebedee Thanks for the feedback! 

The problem was:

    externals: {
      react: 'react'
    }

it should be:

    externals: {
      react: 'React'
    }

I fixed that and also created a [standalone example](https://github.com/moroshko/react-autosuggest/tree/standalone/examples/standalone). Please have a look and let me know if it works for you as expected now.

liamzebedee:
Yeah that works fantastically! Thanks so much for being this fast and helpful :) Have a good one!

moroshko:
Added `dist/standalone` and `examples/standalone` in v1.12.3

PR #25, opened 2015-05-18
Create client-side bundle using browserify

liamzebedee:
This is just an example implementation for reference, so others may read it.


liamzebedee:
#24

moroshko:
Added `dist/standalone` and `examples/standalone` in v1.12.3

PR #26, opened 2015-05-27
Move react to peerDependencies and devDependencies

Halama:
Otherwise multiple versions of React can be loaded which causes weird bugs such as
`Uncaught Error: Invariant Violation: addComponentAsRefTo(...): Only a ReactOwner can have refs. This usually means that you're trying to add a ref to a component that doesn't have an owner (that is, was not created inside of another component's `render` method). Try rendering this component inside of a new top-level component which will hold the ref.`

See also https://github.com/JedWatson/react-select/issues/88 https://github.com/yahoo/flux-router-component/issues/33

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,9 @@
   },
   "dependencies": {
     "classnames": "^2.1.1",
-    "debounce": "^1.0.0",
+    "debounce": "^1.0.0"
+  },
+  "peerDependencies": {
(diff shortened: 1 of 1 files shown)


moroshko:
Thanks a lot for the Pull Request and the relevant links!

moroshko:
It's in v1.14.0 now

PR #32, opened 2015-06-17
Support for controlled components via value prop

### LINKED ISSUE #28: State handling
Hi,
we have encountered issue with component state. The component hold current value of input in it's internal state, so when the property `inputAttributes.value` is changed the change is ignored by Autosuggest component.

Here is the demo https://jsfiddle.net/1y4v9pw6/3/ 

The fix is remove state from component https://github.com/keboola/react-autosuggest/commit/d13008c27e2148205855b0a2fea70dfdc03eaeef and let it behave as controlled component https://facebook.github.io/react/docs/forms.html#controlled-components

Here is the demo same demo but with patched Autosuggest https://jsfiddle.net/2cuf1mzu/1/

This fix BC break, maybe also component which only holds the state could be provided for BC compatiblity.

Our use case where we discovered this issue is form where the initial value of Autosuggest component is dependent on some select value which is used before autosuggest.

Also passing a value instead of event to handler `inputAttributes.onChange` is bit strange.
thanks.

idolize:
This is a breaking change: "value" is no longer a part of inputAttributes and instead is now a top-level property. To support uncontroleld behavior, a new "defaultValue" prop was added.

Fixes #28 
@moroshko

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -7,6 +7,8 @@ import sectionIterator from './sectionIterator';
 
 export default class Autosuggest extends Component { // eslint-disable-line no-shadow
   static propTypes = {
+    value: PropTypes.string,                // Controlled value of the selected suggestion
+    defaultValue: PropTypes.string,         // Initial value of the text
     suggestions: PropTypes.func.isRequired, // Function to get the suggestions
     suggestionRenderer: PropTypes.func,     // Function that renders a given suggestion (must be implemented when suggestions are objects)
     suggestionValue: PropTypes.func,        // Function that maps suggestion object to input value (must be implemented when suggestions are objects)
@@ -16,6 +18,13 @@ export default class Autosuggest extends Component { // eslint-disable-line no-s
     onSuggestionUnfocused: PropTypes.func,  // This function is called when suggestion is unfocused via mouse hover or Up/Down keys
     inputAttributes: PropTypes.object,      // Attributes to pass to the input field (e.g. { id: 'my-input', className: 'sweet autosuggest' })
     id: PropTypes.string                    // Used in aria-* attributes. If multiple Autosuggest's are rendered on a page, they must have unique ids.
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.value) {
+      // If this component is controlled, then handle the value update
+      this.handleValueChange(nextProps.value);
+    }
   }
 
   static defaultProps = {
@@ -32,7 +41,7 @@ export default class Autosuggest extends Component { // eslint-disable-line no-s
 
     this.cache = {};
     this.state = {
-      value: props.inputAttributes.value || '',
+      value: props.value || props.defaultValue || '',
       suggestions: null,
       focusedSectionIndex: null,    // Used when multiple sections are displayed
       focusedSuggestionIndex: null, // Index within a section
@@ -212,7 +221,11 @@ export default class Autosuggest extends Component { // eslint-disable-line no-s
 
   onInputChange(event) {
     const newValue = event.target.value;
+    this.handleValueChange(newValue);
+    this.showSuggestions(newValue);
+  }
 
+  handleValueChange(newValue) {
     this.onSuggestionUnfocused();
     this.onChange(newValue);
 
@@ -220,8 +233,6 @@ export default class Autosuggest extends Component { // eslint-disable-line no-s
       value: newValue,
       valueBeforeUpDown: null
     });
-
-    this.showSuggestions(newValue);
   }
 
   onInputKeyDown(event) {
--- a/README.md
+++ b/README.md
@@ -22,7 +22,8 @@
 * Full control over [styling](#styling) (we just provide the mechanics and classes for you)
 * Full control over [when to show the suggestions](#showWhenOption) (e.g. when user types 2 or more characters)
 * Various hooks: [onSuggestionSelected](#onSuggestionSelectedOption), [onSuggestionFocused](#onSuggestionFocusedOption), [onSuggestionUnfocused](#onSuggestionUnfocusedOption)
-* Ability to [pass props to the input field](#inputAttributesOption) (e.g. initial value, placeholder, onChange, onBlur)
+* Support for [controlled](https://facebook.github.io/react/docs/forms.html#controlled-components) as well as [uncontrolled](https://facebook.github.io/react/docs/forms.html#uncontrolled-components) state
+* Ability to [pass props to the input field](#inputAttributesOption) (e.g. placeholder, onChange, onBlur)
 * In-memory caching (we retrieve suggestions for a given input only once)
 * Thoroughly tested (over 100 tests)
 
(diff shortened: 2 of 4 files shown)


moroshko:
@idolize Thanks a lot for the Pull Request and the updated README!

I have a couple of questions/comments:
* According to React docs for [controlled](https://facebook.github.io/react/docs/forms.html#controlled-components) and [uncontrolled](https://facebook.github.io/react/docs/forms.html#uncontrolled-components) components, `value` and `defaultValue` are `<input>` attributes/props. Why did you decide to take them out of `inputAttributes`?
* If `value` and `defaultValue` live outside of `inputAttributes`, what happens when user specifies both `value` and `inputAttributes.value` (or `defaultValue` and `inputAttributes.defaultValue`)?
* Looks like the [Events playground](http://localhost:3000/examples/dist/index.html#Events playground) example stopped working. My Chrome tab freezes. Does it work for you?
* It would be nice to have tests for the new controlled behaviour.
* As you suggested, we'll have to change the major version. UPGRADE_GUIDE.md would be much appreciated.

Thanks again for your contribution!

idolize:
@moroshko Thanks for taking a look, and for the great feedback. To answer your questions:

- I decided to take `value` and `defaultValue` out of `inputProps` because I think of them more as the actual, selected suggestion value instead of just the transient value of the `<input>` field. When using them in conjunction with  `onSuggestionSelected` this is exactly how it works—the `value` of the `<input>` corresponds to the `this.state.value`, which does not have to equal what is held in the `value` prop.
    - This is just my personal opinion, and I do realize it is a breaking change, so if you would prefer keeping the `value`/`defaultValue` props inside `inputAttributes` then that is totally fine by me :+1: 
- Specifying `value` on the `inputAttributes` does nothing, as the actual `value` prop of the `<input>` is set explicitly to `this.state.value` (which is initialized with the top-level `value`/`defaultValue` props).
    - Perhaps using `PropTypes.shape()` instead of `PropTypes.object` for the propType of `inputAttributes` would make this behavior more clear to the developer, but I didn't want to go that far in case the developer wants to pass in some random props to the `<input>` I hadn't thought of (you are using the spread operator to pass through everything)
- Haven't tested this yet, sorry. I am happy to look at it once I get some more time.
- I agree. I'm still pretty new at writing unit tests for React components (I've mostly just used mocha for pure JS testing), so it may take me a bit longer to add these (feel free to write some tests yourself and I'll try to make them pass if they are not currently)
- If you are still okay with moving `value` and `defaultValue` to the top-level props, I will happily add a UPGRADE_GUIDE.md. Just let me know.
- Will do.

src/Autosuggest.js
```diff
@@ -18,6 +20,10 @@ export default class Autosuggest extends Component { // eslint-disable-line no-s
     id: PropTypes.string                    // Used in aria-* attributes. If multiple Autosuggest's are rendered on a page, they must have unique ids.
   }
 
+  componentWillReceiveProps(nextProps) {
+    this.onUpdate(nextProps.value);
```
idolize: This should be wrapped in an `if (typeof nextProps.value === 'string')` statement so it doesn't trigger if the developer is using the component in an uncontrolled way and they decide to change some other prop (like `inputAttributes.className`). I will update this in my next commit.

idolize:
@moroshko I updated the PR. All feedback should be addressed:

- The Events playground seems to be working fine
- I added some unit tests for the controlled/uncontrolled behavior
- I added a UPGRADE_GUIDE.md
- I got rid of the dist files

Let me know what you think!

dmitry:
I believe `UPGRADE_GUIDE` content is better to put into a readme file.

src/Autosuggest.js
```diff
@@ -18,6 +20,12 @@ export default class Autosuggest extends Component { // eslint-disable-line no-s
     id: PropTypes.string                    // Used in aria-* attributes. If multiple Autosuggest's are rendered on a page, they must have unique ids.
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (typeof nextProps.value === 'string') {
+      this.onUpdate(nextProps.value);
+    }
```
moroshko: If we verify the type here, we shouldn't fail silently. I suggest we have something like:
```
else {
  throw new Error('value must be a string. Found:', nextProps.value);
}
```

src/Autosuggest.js
```diff
@@ -18,6 +20,12 @@ export default class Autosuggest extends Component { // eslint-disable-line no-s
     id: PropTypes.string                    // Used in aria-* attributes. If multiple Autosuggest's are rendered on a page, they must have unique ids.
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (typeof nextProps.value === 'string') {
+      this.onUpdate(nextProps.value);
```
moroshko: What do you think about changing `this.onUpdate()` to `this.updateValue()`? I feel it's a bit more self descriptive this way. (I used `onSomething()` functions in handlers, e.g. `onBlur={this.onInputBlur}`, but here there is no handler.)

moroshko:
@idolize Thanks, looks good!

I think I'm fine with leaving `value` and `defaultValue` on the Autosuggest itself. I just didn't want to create  2 potential places where user could have them (Autosuggest itself and `inputAttributes`) to avoid confusion.

I created a `2.0` branch. Could you please change the Pull Request to `2.0` rather than `master`. I'd like to iterate a bit more on this (add more tests, improve docs, etc) before we merge to `master`.

Thanks a lot for your efforts!

moroshko:
@idolize Also, all the examples in `/examples` are using uncontrolled components. We need to have at least one example for controlled component. Feel free to create one, or let me know how do you think would be the best way to demonstrate the controlled behaviour.

pathsny:
This is pretty awesome. I'm eagerly awaiting this change! One thing I wonder about is what happens when the results are objects rather than strings. Does the initial value get passed to getSuggestionValue to convert the object to a string? or is the "value" prop always a string rather than an object? (I assume it's not).

src/Autosuggest.js
idolize: @moroshko I added this check not to really ensure typing, but to make sure we don't trigger the `onUpdate` method when `props.value` is `null` or `undefined`—because that can cause broken behavior in the uncontrolled case. So, throwing an error here is not a good idea because it will likewise break people trying to use the uncontrolled behavior.

Perhaps this is less confusing?

```js
if (nextProps.value) {
  // If this component is controlled, then handle the value update
  this.onUpdate(nextProps.value);
}
```

src/Autosuggest.js
idolize: Well, this function is still sort of a handler function, because it happens in response to the `value` changing *or* the uncontrolled input text changing due to user typing. But, it is a purely-internal handler function, and not one that can be passed in via props, so perhaps a different naming scheme does make sense.

What do you think about `handleValueChange` as a name?

idolize:
Thanks all!

@pathsny I am not using objects as suggestions currently, so I haven't really investigated that deeply, but from reading the docs it appears the use of objects as options always translates to a string-based value anyway?

idolize:
@moroshko Is there a way to change the target branch of a PR after opening it? I guess I can close this one and open a new one if that is your preference.

pathsny:
@idolize the value object is converted to a string to render using "suggestionValue". I guess the important thing is maintaining the functionality. 

I have an app that needs this, so I can test out the code. But unfortunately your diff only changes the source and not the dist, so I'm confused about how to make it work with my build system which is jspm. But let me try

idolize:
@pathsny Thanks for offering to test! In order to make it work I think you may have to check out the code from my fork, run a build with `npm run build`, create a symlink with `npm link`, and then `npm link react-autosuggest` inside your app's directory. At that point your app should be using the code from this PR.

moroshko:
@pathsny is right. `getSuggestionValue()` is used only in `focusOnSuggestionUsingKeyboard()`. So the initial suggestion object won't go through `getSuggestionValue()`. We need to fix this, and also add tests for a controlled Autosuggest using object suggestions.

@idolize I'm not sure if it's possible to change the target branch in the PR, so let's just create another one and reference this discussion there.

idolize:
Moved to #40

PR #34, opened 2015-06-22
Allow tab key to select suggestion

Jeffrey903:


moroshko:
@Jeffrey903 Thanks for your Pull Request!

What are you trying to achieve?
Currently, if you press TAB, the suggestions list will be closed and the focus will go to the next focusable element, which is normally exactly what is expected from pressing the TAB key (see [bing.com](http://www.bing.com/) and [ebay.com](http://www.ebay.com.au/), for example). Do you want `onSuggestionSelected` to be fired?

[WAI-ARIA guide](http://www.w3.org/TR/wai-aria-practices/#autocomplete) doesn't mention anything about using TAB when interacting with the autosuggest component, so I would prefer to keep the default TAB behaviour.

PR #35, opened 2015-06-24
Bubble event object up to blur listener provided to inputAttributes

mjt01:
When passing an `onBlur` handler through to the `inputAttributes` prop, I would expect that may handler is passed the event object.

I was going to spread the `args` that `onInputBlur` gets passed, but instead followed your pattern of simply passing the `event`, as per the `onInputChange` and `onInputKeyDown` handlers.


moroshko:
@mjt01 Thanks for your contribution!

It's in 1.16.0 now on npm. Enjoy!

PR #37, opened 2015-06-25
Added cache disable prop

CalebMorris:
Still store the results in the cache for use in case the prop is changed without remounting.


CalebMorris:
Need to look through and update to follow your convention of docs

PR #39, opened 2015-06-25
Missing dep

CalebMorris:
Babel change now requires this


moroshko:
@CalebMorris Thanks for the Pull Request!

Could you please provide some pointers on this? Which version of `babel` require `path-is-absolute`? Isn't `path-is-absolute` specified in `babel`'s deps?

CalebMorris:
I'm not sure what version.
When I went to use a fork of your repo, it was failing on the dependency added here not being installed with babel any longer.
I can no longer repo so it may have been a dep of babel. Have been having other problems with it.
Closing unless I get the problem again.

Issue #14, opened 2015-04-22
Clear auto suggest input field


icodejs:
Is there a way to clear the input field after a user has selected a suggestion?

I am interested in what the user has selected, and use this data to update another component but once this is done, I'd like to have access to the input so that I can clear it.

Please can you let me know if this is possible.

Thanks.

moroshko:
@icodejs Currently, there is no nice way to clear the input on selection. You could try something like this:

    class MyComponent extends React.component {
      clearInput() {
        React.findDOMNode(this.refs.autosuggest.refs.input).value = '';
      }
      render() {
        <Autosuggest ref="autosuggest" onSuggestionSelected={this.clearInput} ... />
      }
    }

I haven't tested it, so let me know if it works ;)

I'd be interested to know about your use case. What type of items you have in the autosuggest, and what is that another component that you want to update once suggestion is selected?

icodejs:
Thanks @moroshko,

Your solution worked!!

I had wrapped your auto suggest component in my own pillbox component, which would allow a user to select from the suggestions, and then create a pillbox. Obviously once pillbox was created, the text within the input was no longer be needed.

I will close the issue now as your solution did the trick.

Thanks!

StephanGeorg:
`const autosuggest = this.refs.autosuggest;
autosuggest.props.inputProps.value = '';
autosuggest.input.value = '';
autosuggest.input.focus();`

StephanGeorg:
this.refs.autosuggest.props.inputProps.value = '';
    this.refs.autosuggest.input.value = '';
    this.refs.autosuggest.input.focus();

Serj-M:
> @icodejs Currently, there is no nice way to clear the input on selection. You could try something like this:
> 
> ```
> class MyComponent extends React.component {
>   clearInput() {
>     React.findDOMNode(this.refs.autosuggest.refs.input).value = '';
>   }
>   render() {
>     <Autosuggest ref="autosuggest" onSuggestionSelected={this.clearInput} ... />
>   }
> }
> ```
> 
> I haven't tested it, so let me know if it works ;)
> 
> I'd be interested to know about your use case. What type of items you have in the autosuggest, and what is that another component that you want to update once suggestion is selected?

I came across a similar task.
Changes to other components must clear the input autosuggest. How to do it? I am a beginner React programmer.
I tried your code, but I get a lot of errors. Perhaps there is not placing your code. Please show a working code example.
[...]

PR #45, opened 2015-07-01
Add onInputChange callback

tpievila:
Add a callback for input so that the parent can keep track of the input state and fully control the autosuggest component. Part of solving #28.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,30 @@ function onSuggestionUnfocused(suggestion) { // In this example 'suggestion' is 
              onSuggestionUnfocused={onSuggestionUnfocused} />
 ```
 
+<a name="onInputChange"></a>
+#### onInputChange (optional)
+
+This function will be called when user changes the input.
+Combined with [value](#valueOption) you can implement a fully
+[controlled](https://facebook.github.io/react/docs/forms.html#controlled-components)
+component.
+
+```js
+function(input) {
+  ...
+}
+```
+
+* `input` - Will be the value of the input field
+
+For example:
+
+```js
+function onInputChange(input) {
+  console.log('Got new input: [' + input + ']');
+}
+```
+
 <a name="inputAttributesOption"></a>
 #### inputAttributes (optional)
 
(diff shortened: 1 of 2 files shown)


moroshko:
@tpievila Thanks for the Pull Request!

I wonder what `onInputChange` gives us that `inputAttributes.onChange` doesn't?

tpievila:
onInputChange fires only when the user types something to the field. onChange fires also on mouse and keyup/keydown. I suppose onChange could also get more information on what kind of interaction actually happened?

moroshko:
@tpievila If `onInputChange` is not fired on every input change, then the name is misleading. Input value can also change if user hits ESC (this clears the input), or if user pastes text from clipboard (Cmd + V).

I agree, `inputAttributes.onChange` could provide more information about the last interaction.

Could you explain please what are you trying to achieve?

tpievila:
I'm tracking the value state in the parent so that I can change it based on other interactions. Currently I'm figuring out how to best have two fields (to and from) where the contents can be switched with a button (switch direction). If I tracked also the keyup/down event changes, then the user would effectively select a choice when clicking the switch button even without having confirmed the selection by enter.

Certainly an edge case, but didn't come up with a better solution so far.

moroshko:
@tpievila I'm still a bit confused. Here is what I understand so far:

* There are 2 autosuggests: `from` and `to`
* There is a switch button. When clicked, the values of `from` and `to` should switch.

(This reminds me Google Maps.)

What do you want to do when input's value is changed?

tpievila:
Merely to track the changes, not yet react.

But in general, if the parent can force the value of the input field with the new value option, it needs to track it. I was thinking that the keyboard changes should be ignored before actual selection, but true, the new callback in this PR doesn't handle clearing with ESC.

The problem with onChange firing in so many places is that I get into an endless loop when I track the state in parent and feed it back in the value option/prop.

moroshko:
@tpievila Not exactly sure what you mean by "track". With `inputAttributes.onChange` you know when the value changes. In version 2.0 you'll get the `event` as well, so you'll have a better control on handling changes. Could you explain **what** are you trying to achieve, without telling the **how**?

tpievila:
Switching the field contents. Not sure how to clarify it. Based on your previous message, I thought that the what was clear and you were asking for how.

moroshko:
@tpievila So you basically want to set input's value without `onChange` being called, or alternatively know if `onChange` is called as a result of setting input's value?

tpievila:
Indeed, yes.

Another case I'm now trying to figure out is how to combine this with clearing the value completely after a suggestion has been selected. The trick at #14 doesn't work when the value prop is used to control the component.

moroshko:
@tpievila I believe you'll be able to achieve what you want if we add a "source" parameter to `onChange`.

PR #54, opened 2015-07-30
Support CSS Modules and inline styles for theming

markdalgleish:
This allows themes to be provided following the [react-themeable](https://github.com/markdalgleish/react-themeable) standard.

The default theme values match the existing global classes defined in your readme, so this maintains support for those still using global CSS. Of course, these values now form part of your component's public API, so you may want to rename them.

This code in this PR is undocumented. I attempted to update the readme but it turned out to be a much larger change than I anticipated, so I figured it's best left up to you. I'm happy to help, of course.


dmitry:
It doesn't follow the BEM standard: block, element, modifier? It only allows to set block and element?

krvital:
+1 for css modules

PR #70, opened 2015-10-01
add focusAfterSuggestionSelected optional prop, default true

hughlomas:
I'm running into a problem on mobile devices where I want to blur the input after a selection, mainly so that the device's on screen keyboard will hide after an option is picked. 

If I blur in the `onSuggestionSelected` event callback handler, the `findDOMNode(this.refs.input).focus();` in `onSuggestionMouseDown` ends up focusing the input again. 

Right now I've resorted to using a timeout in `onSuggestionSelected` to blur the input. I'd prefer a simple option to control the behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Check out the [standalone example](https://github.com/moroshko/react-autosuggest
 * [`id`](#idOption)
 * [`scrollBar`](#scrollBarOption)
 * [`theme`](#themeOption)
+* [`focusAfterSuggestionSelected`](#focusAfterSuggestionSelected)
 
 <a name="suggestionsOption"></a>
 #### suggestions (required)
@@ -451,6 +452,15 @@ The following diagrams explain the classes above.
     |  +----------------------------------------------------------------------+  |
     |                                                                            |
     +----------------------------------------------------------------------------+
+
+<a name="focusAfterSuggestionSelectedOption"></a>
+#### focusAfterSuggestionSelected (optional)
+
+Defaults to `true`
+
+Input will receive focus after a suggestion is selected
+
+Set `focusAfterSuggestionSelected={false}` to disable this behaviour.
 
 ## Development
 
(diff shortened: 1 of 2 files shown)


r3to:
I have the same problem. Thanks @hughlomas for your PR

moroshko:
@hughlomas Sorry it took me so long to get back to you. It totally make sense to hide the keyboard on mobile devices after selection. The only issue I can see with this approach is that it will affect desktop users as well, i.e. if you set `focusAfterSuggestionSelected={false}`, desktop users won't get the focus after selection. Does this mean we need to handle mobile-only behaviour separately? What are your thoughts on this? cc @r3to

r3to:
hi @moroshko! In my use case, I want to remove the focus on desktop and mobile devices after the user has selected one of the suggestions. I guess it would be more intuitive to just have a `focusAfterSuggestionSelected` option. So, if the behavior on mobile and desktop should differ, this could be easily changed per configuration: `focusAfterSuggestionSelected={isDesktop}`

moroshko:
@r3to It all makes sense. I'm just curious:
1. why are you not interested in having the focus back on desktop when suggestion is selected?
2. how would you implement `isDesktop`?

r3to:
@moroshko 
1. I'm building a meteorological site, where the Input field for the location also acts as the title for the current forecast. If a valid location is entered/selected, the input field will be styled slightly different, than when it's foucssed. I'll send you a link to the website, once we go live :-)
2. I'd probably use a user agent sniffing method to define `isDesktop`, knowing that this isn't a very stable and future-proof way. I'd probably use https://github.com/kaimallea/isMobile for that (haven't used it yet). I think `isDesktop` shouldn't be defined by react-autosuggest, but by the component, that uses `Autosuggest``

oh, and btw: thanks for your work!

moroshko:
@r3to Thanks!!

@hughlomas Could you please resolve the conflicts so that I could merge this PR?

moroshko:
@hughlomas @r3to I just added a `focusInputOnSuggestionClick` prop to 3.0.0-beta.2. You are welcome to upgrade to 3.0:

hughlomas:
Sorry I missed your messages, @moroshko, I am not a super active github user and it had been a while; I forgot about this. Thank you for merging it in!

Issue #73, opened 2015-10-14
Can't scroll autosuggestion list on mobile device


hannesj:
[This](https://github.com/moroshko/react-autosuggest/blob/v2.1.0/src/Autosuggest.js#L514) makes it possible to get faster feedback on mobile devices, but unfortunately it makes it impossible to scroll the list, as the onTouchStart fires also when starting to drag the list.

moroshko:
@hannesj `onTouchStart` was added because iOS devices didn't pick up `onMouseDown` at all, so tapping a suggestion on iPhone, for example, was causing the suggestions list to close without updating the input value. That's the history behind `onTouchStart`.

Any ideas how to fix this? Pull request is welcome!

bachstatter:
Hi! 
I made a PR using [react-tap-event-plugin]  (https://github.com/zilverline/react-tap-event-plugin). I added some thoughts as well. Please let me know what you think.

moroshko:
Thanks a lot! Version 2.1.1 is on npm now

PR #75, opened 2015-10-14
ISSUE #73 Use touchTap event instead of touchstart.

bachstatter:
So we need a tapEvent since react doesnt support it (yet). It's possible to do a custom solution that saves touchStarts position and a timestamp. Then onTouchEnd compare that position and check how long time that has passed. This way we can ignore scrolling and "long touches".

A faster and more clean solution is the [react-tap-event-plugin] (https://github.com/zilverline/react-tap-event-plugin/blob/master/README.md). It hooks into reacts event system and adds a touchTap event.
Then when react adds touchTap event its easy to remove this dependency.

From react-tap-event-plugin:
You've probably heard of [iOS's dreaded 300ms tap delay](http://updates.html5rocks.com/2013/12/300ms-tap-delay-gone-away).  React's `onClick` attribute falls prey to it.  Facebook's working on a solution in the form of `TapEventPlugin`, but it [won't be made available](https://github.com/facebook/react/issues/436) [until 1.0](https://github.com/facebook/react/pull/1170).

If you're reading this, you're probably working on a project that can't wait until they figure out how they want to publish it.  This repo is for you.

When Facebook solves [#436](https://github.com/facebook/react/issues/436) and [#1170](https://github.com/facebook/react/pull/1170), this repo will disappear.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "debounce": "^1.0.0",
+    "react-tap-event-plugin": "^0.2.1",
     "react-themeable": "^1.0.1"
   },
   "peerDependencies": {
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -3,6 +3,8 @@ import { findDOMNode } from 'react-dom';
 import debounce from 'debounce';
 import themeable from 'react-themeable';
 import sectionIterator from './sectionIterator';
+import injectTapEventPlugin from 'react-tap-event-plugin';
+injectTapEventPlugin();
 
 export default class Autosuggest extends Component {
   static propTypes = {
@@ -511,7 +513,7 @@ export default class Autosuggest extends Component {
             onMouseEnter={() => this.onSuggestionMouseEnter(sectionIndex, suggestionIndex)}
             onMouseLeave={() => this.onSuggestionMouseLeave(sectionIndex, suggestionIndex)}
             onMouseDown={onSuggestionClick}
-            onTouchStart={onSuggestionClick}>
+            onTouchTap={onSuggestionClick}>
           {this.renderSuggestionContent(suggestion)}
         </li>
       );


bachstatter:
If you want to try it out on a phone you could change server.js to:
`}).listen(3000, 'localhost', function(err, result) {`

`}).listen(3000, '', function(err, result) {`

This way you can surf to your ip instead of using localhost. This is also possible to do from your phone as long as you are on the same network :)

moroshko:
Thanks a lot!

moroshko:
@bachstatter Including `react-tap-event-plugin` in `react-autosuggest` presents a challenge: If another 3rd party component or the app itself uses `react-tap-event-plugin`, we'll get:
> Uncaught Error: Invariant Violation: EventPluginRegistry: Cannot inject two different event plugins using the same name, `TapEventPlugin`.

See: https://github.com/zilverline/react-tap-event-plugin/issues/47

What are your thoughts about **not including** `react-tap-event-plugin` as part of `react-autosuggest`, and make it clear in the readme that it needs to be included manually to support mobile devices? Alternatively, we could add a config option in `react-autosuggest` to conditionally include `react-tap-event-plugin`. Or, maybe you have a better idea?

bachstatter:
Thats an interesting challenge. I'm not sure if the best solution is to add a config option or just some instructions in the readme.

I found a third solution, we can find all events in the EventPluginRegistry. This if statement is false before injectTapEventPlugin() is invoked and true after. I'm not sure how reliable this is, but it seems to work.

```javascript
import { registrationNameModules } from 'react/lib/EventPluginRegistry';

if (!registrationNameModules.onTouchTap) {
  injectTapEventPlugin();
}
```
What do you think? We could wait and see if zilverline respond to your issue.

moroshko:
@bachstatter Checking `registrationNameModules.onTouchTap` is problematic, I think. If some other library is doing something with `onTouchTap`, and it run first, our code will do nothing. Or, if `react-autosuggest` runs first, we'll break the other plugin functionality.

bachstatter:
The first case should be fine, since if onTouchTap is injected by some other library we wouldn't need to do it. But it might break other projected trying to use onTouchTap.  

We could use https://github.com/JedWatson/react-tappable and wrap each item with the a Tappable component.

I'm not sure which solution is best...

Issue #76, opened 2015-10-20
configure debounce delay


loopmode:
It would be great if we could configure the debounce delay time. Currently, it is hard-coded as 100ms, in Autosuggest.js ~line 63:

    this.suggestionsFn = debounce(props.suggestions, 100);

This can generate unnecessary load on the servers and I personally would like to set it to ~500ms for our particular project.

I can create a PR if you don't have time for this, but consider it as useful as I do.

moroshko:
@loopmode I agree, it needs to be configurable. PR is welcome!

loopmode:
basically an easy task, kind of finished. I struggle with getting an example running and line endings tho :) Hope i find time for a proper PR soon

moroshko:
@loopmode In 3.0, debounce is not part of react-autosuggest anymore. Check out [this example](https://github.com/moroshko/react-autosuggest/tree/3.0/demo/src/components/App/components/DebouncedExample) to see how to implement debouncing.

    npm install react-autosuggest@3.0-latest --save

avindra:
@moroshko : Your link does not work. Is there a way to debounce in the latest version?

avindra:
To answer my own question, I was able to use `lodash` to debounce:

```
const debounce = require('lodash.debounce');

...
    this.onSuggestionsUpdateRequested = debounce(this.onSuggestionsUpdateRequested.bind(this), 300);
```

Something like that should work for you :+1:

moroshko:
@avindra Sorry for the broken link, I fixed that. Have a look at [this example](http://codepen.io/moroshko/pen/KVaGJE).

maksimgm:
@moroshko Really useful. Thank you for the demo link. You're a life saver :)

muellermarkus:
Note that for autosuggestions, lodash's **`_.throttle` might often be a better fit** instead of `_.debounce`. 

`debounce` will wait with invoking `this.onSuggestionsUpdateRequested` until the user has stopped typing. If a user is typing a long query, **he will only get auto-suggestions when he pauses typing or has finished typing**. => So that `wait` milliseconds have passed since the last keystroke.

What you more likely want is the that `this.onSuggestionsUpdateRequested` does **get invoked while the user is typing, but not too often**, i.e. not on every keystroke. The `throttle` method offers exactly that. `this.onSuggestionsUpdateRequested` will get invoked at most every `wait` milliseconds. 

[...]

PR #80, opened 2015-10-26
Added debounce prop

loopmode:
Allows custom debounce delays for fetching and showing suggestions. The default value of 100ms is still used, however, the component now accepts the optional 'debounce' prop / attribute.

Added short explanation in the readme file.
Did not add separate example.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -22,7 +22,8 @@ export default class Autosuggest extends Component {
     cache: PropTypes.bool,                  // Set it to false to disable in-memory caching
     id: PropTypes.string,                   // Used in aria-* attributes. If multiple Autosuggest's are rendered on a page, they must have unique ids.
     scrollBar: PropTypes.bool,              // Set it to true when the suggestions container can have a scroll bar
-    theme: PropTypes.object                 // Custom theme. See: https://github.com/markdalgleish/react-themeable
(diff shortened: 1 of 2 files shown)


src/Autosuggest.js
```diff
@@ -90,6 +92,15 @@ export default class Autosuggest extends Component {
         this.handleValueChange(nextProps.value);
       }
     }
+    if (nextProps.debounce !== this.props.debounce || nextProps.suggestions !== this.props.suggestions) {
+      this.updateSuggestionsFn(nextProps);
+    }
```
moroshko: At the moment, the only value that can change after the initialization is `value` (this is the controlled behaviour). Changing others will be supported in 3.0, so I suggest we don't update `this.suggestionsFn` on the fly. Would you be happy to remove this for now? Also, could you please squash your commits into a single commit?

loopmode:
Sure, will do!

moroshko:
@loopmode Just wanted to let you know that 3.0 alpha version is available on npm:

    npm install react-autosuggest@3.0.0-alpha.1 --save

moroshko:
Closing as debouncing is not part of 3.0 anymore. Check out [this example](https://github.com/moroshko/react-autosuggest/tree/3.0/demo/src/components/App/components/DebouncedExample) to see how to implement custom debouncing.

avindra:
FYI see #76 for working examples

PR #86, opened 2015-11-03
Satisfy react-dom by react.js in standalone config

bendlas:
Right now, the webpack standalone config unnessecarily includes react-dom, which bloats unminified size to 640K. With this change, react-dom is satisfied by `window.React`, which shrinks unminified size down to ~105K.

### DIFF
--- a/webpack.standalone.config.js
+++ b/webpack.standalone.config.js
@@ -21,7 +21,8 @@ module.exports = [{
   },
 
   externals: {
-    react: 'React'
+    'react': 'React',
+    'react-dom': 'React'
   }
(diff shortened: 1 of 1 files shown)


moroshko:
@bendlas Thanks for your PR. Sorry it took me so long to get back to you.

`/dist/standalone/autosuggest.js` is just 35.4 kB in version 2.2.5, and it doesn't get smaller if I add `'react-dom': 'React'` in `webpack.standalone.config.js`. I believe this is what you wanted to achieve, but happy to reopen the issue if this is not the case.

PR #87, opened 2015-11-03
Reset callback fns when props are updated

bendlas:
After updating an Autosuggest widget with a new suggestions fn, it still uses the previous fn, because it only saves the suggestion fn in the constructor. The attached patch resets `suggestions`, as well as `onFocus`, `onChange` and `onBlur` from `componentWillReceiveProps`. Maybe more state needs to be reset in order to guarantee correctness.


moroshko:
@bendlas Thanks for your PR. This shouldn't be a problem in 3.0 anymore. You are welcome to try it, and report back if it's still an issue.

PR #94, opened 2015-12-08
onSuggestionSelected return null suggstion, in korean(onChange event …

nothinking:
In korean, japanese and chinese. (different encoding? with english)
onChange event tirggered at MouseClick.

So, at onSuggestSelected, recieve null suggstion.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -106,9 +106,6 @@ export default class Autosuggest extends Component {
     this.resetSectionIterator(suggestions);
     this.setState({
       suggestions: suggestions,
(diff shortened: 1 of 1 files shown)


moroshko:
@nothinking Sorry, I'm not sure I understand the issue you are fixing. Could you elaborate please, or provide an example that demonstrates the issue?

moroshko:
If this is still an issue, please try to reproduce with react-autosuggest 3.0. You can start with [this example](http://codepen.io/moroshko/pen/wMzQKV).

Happy to reopen the issue if a reproducible example is provided.

PR #95, opened 2015-12-08
Add escape hatch to avoid conflict with FastClick

builtinnya:
When using with [FastClick] on affected mobile devices/browsers,  clicking a suggestion just fires `blur` event but not `mousedown` event. It just hides the suggestion list and the input remains unchanged.

This PR adds `ignoredByFastClick` option, which, if set to `true`, makes [FastClick] ignore suggestion elements. `ignoredByFastClick` defaults to `false` to avoid compatibility issues.

I'm afraid that I can't come up with automated tests for this. ([FastClick] also does not have automated tests.)

To reproduce the problem, just load and initialize [FastClick] on `examples/dist/index.html`, then input some text and click a suggestion element on an affected device or chrome dev tool mobile simulator.

[FastClick]: https://github.com/ftlabs/fastclick

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ Check out the [standalone example](https://github.com/moroshko/react-autosuggest
 * [`id`](#idOption)
 * [`scrollBar`](#scrollBarOption)
 * [`theme`](#themeOption)
+* [`ignoredByFastClick`](#ignoredByFastClickOption)
 
 <a name="suggestionsOption"></a>
 #### suggestions (required)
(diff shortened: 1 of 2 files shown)


moroshko:
@builtinnya Could you please check if 3.0 still has the issue?

3.0 uses `onClick` instead of `onMouseDown`, so I suspect that the issue may be resolved there.

moroshko:
@builtinnya I just tested version 3.0.0-beta.3 on iOS and everything looks ok even when FastClick is enabled.

builtinnya:
@moroshko 
Sorry for being late... and thank you for your kindness.
I tested version 3.0.0-beta.4 on Chrome dev tool mobile simulator and confirmed that it's been fixed.
I wish you a Happy New Year!

moroshko:
No worries. Happy New Year :fireworks:

PR #111, opened 2016-01-20
change this.props.theme to props.theme, fixes #110

### LINKED ISSUE #110: Fails in IE10
First of all, your component is great. I noticed though it fails in IE10 with this error:
```
Unable to get property 'theme' of undefined or null reference
```

Seems like `this.props` is `undefined` [here](https://github.com/moroshko/react-autosuggest/blob/master/src/AutosuggestContainer.js#L106).

cheapsteak:
Ran through tests and linting with no issues

Fixes
https://github.com/moroshko/react-autosuggest/issues/110


moroshko:
Thanks!

PR #132, opened 2016-03-04
Add optional maxLength prop

bricejlin:


moroshko:
@bricejlin Thanks for your Pull Request!

Limiting the number of suggestions is a very common requirement. All you have to do is to add `.slice(0, limit)` in the function that returns the suggestions in your app.

For example, [I modified the Basic Usage example](http://codepen.io/moroshko/pen/WwQzrB):



I'd like to keep the Autosuggest API as simple and flexible as possible, and adding an extra prop (`maxLength`) doesn't feel right.

bricejlin:
@moroshko ah I see. I will try that then. I was afraid that would cut off the overall list from which Autosuggest filters through. Whereas my usecase was for Autosuggest to have the entire list but only show a limited result list.

moroshko:
@bricejlin Note that Autosuggest gets only `suggestions`, which is the "filtered list". You never pass the "entire list" to it.

akhayoon:
@moroshko Let me know if this is still your view. I would happily do a PR if you like. Cheers

moroshko:
@akhayoon This is still my view :)

PR #145, opened 2016-04-11
Update package.json for react & postcss-loader

SwaroopH:


package.json
```diff
@@ -28,10 +28,10 @@
     "redux": "^3.3.1"
   },
   "peerDependencies": {
-    "react": "^0.14.7 || ^15.0.0-rc.1"
+    "react": "^0.14.7 || ^15.0.1"
   },
   "devDependencies": {
-    "autoprefixer-loader": "^3.2.0",
+    "postcss-loader": "^0.8.2",
```
moroshko: Maybe I'm missing something, but shouldn't we also update Webpack configs? Where is `postcss-loader` used?

moroshko:
This fix is in master now

SwaroopH:
@moroshko sorry I missed your previous comment - need a better way to filter my github emails!

PR #153, opened 2016-04-26
Suppress focusing with mouse hover while focusing with keyboard.

pdc:
This is needed to avoid a UI glitch when setting suggestions list has a scrollbar (because it has max-height and overflow style properties). On the demo page the languages sample has had 100 extra entries added to demonstrate scrolling.

To see the issue it is supposed to address:
1. Enter `Z` to show 100 suggestions.
2. Use arrow keys to set the focused selection
3. Move mouse cursor over the suggestion list (this should have no immediate effect)
4. Use keys to move focus down until the suggestion list scrolls

When I do this in Chrome, the suggestion that falls under the mouse cursor because of the scrolling gets the focus, rather than the intended one.

The way this change attempts to fix this is by suppressing the setting of focus on hover immediately following change of focus via the keyboard.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -1,6 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
-import { inputFocused, inputBlurred, inputChanged, updateFocusedSuggestion,
+import { inputFocused, inputBlurred, inputChanged, updateFocusedSuggestion, updateFocusedUsingKeyboard,
          revealSuggestions, closeSuggestions } from './reducerAndActions';
 import Autowhatever from 'react-autowhatever';
 
@@ -10,6 +10,7 @@ function mapStateToProps(state) {
     isCollapsed: state.isCollapsed,
     focusedSectionIndex: state.focusedSectionIndex,
     focusedSuggestionIndex: state.focusedSuggestionIndex,
+    focusedUsingKeyboard: state.focusedUsingKeyboard,
     valueBeforeUpDown: state.valueBeforeUpDown,
     lastAction: state.lastAction
   };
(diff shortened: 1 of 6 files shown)


moroshko:
@pdc Could you please describe the issue that this PR addresses?

pdc:
Thanks! I've rebased the code and expanded the description a bit.

talarari:
looks like this is ready for release?

src/Autosuggest.js
```diff
@@ -274,10 +280,19 @@ class Autosuggest extends Component {
...
     const onMouseLeave = () => {
-      updateFocusedSuggestion(null, null);
+      if (!focusedUsingKeyboard) {
+        updateFocusedSuggestion(null, null);
+      }
+    };
+    const onMouseMove = () => {
+      if (focusedUsingKeyboard) {
+        updateFocusedUsingKeyboard(false);
+      }
```
moroshko: I wonder if we can avoid having the `onMouseMove` handler. This handler would potentially be called hundreds of times if user "reviews" the suggestions by moving the mouse downwards. By introducing `onMouseMove` we penalise all users that don't have a scrollbar.

moroshko:
@pdc This is a tricky issue to solve, so thanks a lot for your time and effort!

src/Autosuggest.js
moroshko: I wonder if running `updateFocusedUsingKeyboard(false)` in a `setTimeout` after `updateFocusedSuggestion(newFocusedSectionIndex, newFocusedItemIndex, value, true);` would work?
Just looking for ways to avoid adding `onMouseMove`.
What do you think?

pdc: I tried something like this earlier and had trouble with trhe enter/leave events arriving after the timeout. I could try it with a longer delay 

Another option might be to add the mouysemove handler only when focus has been set with the keyboard and remove the handler the firsst time it is called.

I'll have a go with these and update the PR if I get one of them to work :-)

pdc:
Turns out React makes switching the mouse-movement event-handling on only when you have focused using the keyboard easy :grinning:

moroshko:
@pdc Could you please rebase on master to resolve conflicts?

moroshko:
@pdc ping

moroshko:
Closing for now.

PR #155, opened 2016-04-26
Always render suggestions – fix #133

### LINKED ISSUE #133: Always render suggestions
Hey hey,

I have a specific use case where I need the list of suggestions to always be rendered. I expected this to be the case if [`shouldRenderSuggestions`](https://github.com/moroshko/react-autosuggest#shouldrendersuggestions-optional) always returned true, but things are a bit more complicated.

In reality, for suggestions to be rendered the criteria are:

- `isFocused`, and
- `!isCollapsed`, and
- `suggestions.length > 0`, and
- Result from `shouldRenderSuggestions` 

(see https://github.com/moroshko/react-autosuggest/blob/master/src/Autosuggest.js#L177)

I think it makes perfect sense to check for "are there suggestions" and I don't know what `isCollapsed` is about. For `isFocused` however, this seems like a sensible default but not necessarily something that this component should have opinions about. `react-autosuggest` strikes me as rather unopinionated in other parts of its API. 

I can see at least three ways to tackle this:

- Have a configuration option to define whether `isFocused` (and `isCollapsed`?) should be checked. This seems like the simplest choice because it's fully backward compatible, but not the best IMHO because `shouldRenderSuggestions`'s name suggests that it should be responsible for this.
- Pass `isFocused` to `shouldRenderSuggestions`, and let the component's user decide whether they want to take it into account or not. Seems like the option that aligns best with the existing API. It's just an extra parameter, and its logic is somewhat similar to that of giving a `reason` for `onSuggestionsUpdateRequested` and leaving it up to the user to decide what they want to do. However this isn't backwards compatible.
- If you don't think this use case is worth supporting, then at least update [`shouldRenderSuggestions`](https://github.com/moroshko/react-autosuggest#shouldrendersuggestions-optional)'s documentation to make it clear that suggestions are only rendered _when input field is focused_.

In those three cases I'm happy to contribute a PR, but I think this is worth discussing in advance :)

ThibWeb:
Hey hey,

this is a PR to fix #133. It implements a new `alwaysRenderSuggestions` boolean prop, as we discussed. When `alwaysRenderSuggestions={true}`, the checks to decide whether to render the suggestions are bypassed. Also comes with some documentation on this new prop, and relevant tests.

I'd be keen to get a review of whether [the tests are relevant and thorough enough](https://github.com/ThibWeb/react-autosuggest/blob/feature/always-render-suggestions/test/plain-list/AutosuggestApp.test.js#L431), and make sure this new prop [is used in the right places](https://github.com/moroshko/react-autosuggest/pull/155/commits/0b0a86cebe3391adabf995bb02323ed478381877).

Also two things occurred to me while writing this PR:

- Having an example for this would only be relevant if it came with the kind of UI this is useful for. Do we want to make one? (another way to see this would be to have a list of "projects `react-autosuggest` is used on" somewhere).
- It might or might not be be obvious that in order for this to be useful, you need to have a list of suggestions to display _even when the user hasn't typed anything yet_ (otherwise there's nothing to display, doh). Does this need to be stated in the docs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ class Example extends React.Component {
 * [`renderSuggestion`](#renderSuggestionProp)
 * [`inputProps`](#inputPropsProp)
 * [`shouldRenderSuggestions`](#shouldRenderSuggestionsProp)
+* [`alwaysRenderSuggestions`](#alwaysRenderSuggestionsProp)
 * [`multiSection`](#multiSectionProp)
 * [`renderSectionTitle`](#renderSectionTitleProp)
 * [`getSectionSuggestions`](#getSectionSuggestionsProp)
@@ -330,6 +331,15 @@ function shouldRenderSuggestions(value) {
   return value.trim().length > 2;
 }
 ```
+
+When `shouldRenderSuggestions` returns true, suggestions will be rendered only when the input field is focused. If you would like to render suggestions regardless of whether the input field is focused or not, set `alwaysRenderSuggestions={true}` (`shouldRenderSuggestions` is ignored in this case).
+
+<a name="alwaysRenderSuggestionsProp"></a>
+#### alwaysRenderSuggestions (optional)
+
+By default, suggestions are rendered only if the input field is focused.
+
+If you'd like to _always_ display suggestions, set `alwaysRenderSuggestions={true}`.
 
 <a name="multiSectionProp"></a>
 #### multiSection (optional)
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -22,6 +22,7 @@ class Autosuggest extends Component {
     renderSuggestion: PropTypes.func.isRequired,
     inputProps: PropTypes.object.isRequired,
     shouldRenderSuggestions: PropTypes.func.isRequired,
+    alwaysRenderSuggestions: PropTypes.bool.isRequired,
     onSuggestionSelected: PropTypes.func.isRequired,
     multiSection: PropTypes.bool.isRequired,
     renderSectionTitle: PropTypes.func.isRequired,
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -61,14 +61,16 @@ class Autosuggest extends Component {
 
   componentWillReceiveProps(nextProps) {
     if (nextProps.suggestions !== this.props.suggestions) {
-      const { suggestions, inputProps, shouldRenderSuggestions,
-              isCollapsed, revealSuggestions, lastAction } = nextProps;
+      const {
+        suggestions, inputProps, shouldRenderSuggestions, isCollapsed,
+        revealSuggestions, lastAction, alwaysRenderSuggestions
+      } = nextProps;
       const { value } = inputProps;
 
-      if (suggestions.length > 0 && shouldRenderSuggestions(value)) {
+      if (alwaysRenderSuggestions || suggestions.length > 0 && shouldRenderSuggestions(value)) {
         this.maybeFocusFirstSuggestion();
 
-        if (isCollapsed && lastAction !== 'click' && lastAction !== 'enter') {
+        if (alwaysRenderSuggestions || isCollapsed && lastAction !== 'click' && lastAction !== 'enter') {
           revealSuggestions();
         }
       }
@@ -226,10 +228,10 @@ class Autosuggest extends Component {
       theme, isFocused, isCollapsed, focusedSectionIndex, focusedSuggestionIndex,
       valueBeforeUpDown, inputFocused, inputBlurred, inputChanged,
       updateFocusedSuggestion, revealSuggestions, closeSuggestions,
-      getSuggestionValue
+      getSuggestionValue, alwaysRenderSuggestions
     } = this.props;
     const { value, onBlur, onFocus, onKeyDown } = inputProps;
-    const isOpen = isFocused && !isCollapsed && this.willRenderSuggestions();
+    const isOpen = alwaysRenderSuggestions || isFocused && !isCollapsed && this.willRenderSuggestions();
     const items = (isOpen ? suggestions : []);
     const autowhateverInputProps = {
       ...inputProps,
--- a/test/plain-list/AutosuggestApp.js
+++ b/test/plain-list/AutosuggestApp.js
@@ -48,9 +48,11 @@ export const shouldRenderSuggestions = sinon.spy(value => {
   return value.trim().length > 0 && value[0] !== ' ';
 });
 
-export const toggleAlwaysRenderSuggestions = sinon.spy((toggle, callback) => {
+export const setAlwaysRenderSuggestions = sinon.spy((toggle, callback) => {
   app.setState({
-    alwaysRenderSuggestions: toggle
+    alwaysRenderSuggestions: toggle,
+    // Using "elm" to retrieve suggestions even when input is blank.
+    suggestions: getMatchingLanguages(app.state.value || 'elm')
   }, callback);
 });
 
--- a/test/plain-list/AutosuggestApp.js
+++ b/test/plain-list/AutosuggestApp.js
@@ -48,6 +48,12 @@ export const shouldRenderSuggestions = sinon.spy(value => {
   return value.trim().length > 0 && value[0] !== ' ';
 });
 
+export const toggleAlwaysRenderSuggestions = sinon.spy((toggle, callback) => {
+  app.setState({
+    alwaysRenderSuggestions: toggle
+  }, callback);
+});
+
 export const onSuggestionsUpdateRequested = sinon.spy(({ value }) => {
   app.setState({
     suggestions: getMatchingLanguages(value)
@@ -64,7 +70,8 @@ export default class AutosuggestApp extends Component {
 
     this.state = {
       value: '',
-      suggestions: getMatchingLanguages('')
+      suggestions: getMatchingLanguages(''),
+      alwaysRenderSuggestions: false
     };
   }
 
@@ -75,7 +82,7 @@ export default class AutosuggestApp extends Component {
   }
 
   render() {
-    const { value, suggestions } = this.state;
+    const { value, suggestions, alwaysRenderSuggestions } = this.state;
     const inputProps = {
       id: 'my-awesome-autosuggest',
       placeholder: 'Type a programming language',
@@ -93,6 +100,7 @@ export default class AutosuggestApp extends Component {
         renderSuggestion={renderSuggestion}
         inputProps={inputProps}
         shouldRenderSuggestions={shouldRenderSuggestions}
+        alwaysRenderSuggestions={alwaysRenderSuggestions}
         onSuggestionSelected={onSuggestionSelected}
         ref={this.storeAutosuggestReference} />
     );
--- a/demo/src/components/App/components/Examples/Examples.js
+++ b/demo/src/components/App/components/Examples/Examples.js
@@ -4,6 +4,7 @@ import React from 'react';
 import Basic from 'Basic/Basic';
 import MultipleSections from 'MultipleSections/MultipleSections';
 import CustomRender from 'CustomRender/CustomRender';
+import AlwaysOpen from 'AlwaysOpen/AlwaysOpen';
 
 export default function Examples() {
   return (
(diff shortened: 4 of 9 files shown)


test/plain-list/AutosuggestApp.js
```diff
@@ -48,6 +48,13 @@ export const shouldRenderSuggestions = sinon.spy(value => {
   return value.trim().length > 0 && value[0] !== ' ';
 });
 
+export const toggleAlwaysRenderSuggestions = sinon.spy((toggle, callback) => {
+  app.setState({
+    alwaysRenderSuggestions: toggle,
+    suggestions: getMatchingLanguages(app.state.value || 'elm')
```
moroshko: 👍 for "elm", but why `|| 'elm'` is required?

src/Autosuggest.js
```diff
@@ -176,13 +179,14 @@ class Autosuggest extends Component {
     const {
       suggestions, renderSuggestion, inputProps, shouldRenderSuggestions,
       onSuggestionSelected, multiSection, renderSectionTitle, id,
-      getSectionSuggestions, focusInputOnSuggestionClick, theme, isFocused,
+      getSectionSuggestions, focusInputOnSuggestionClick, theme,isFocused,
```
moroshko: really?

README.md
```diff
@@ -328,6 +329,13 @@ function shouldRenderSuggestions(value) {
 }
 ```
 
+<a name="alwaysRenderSuggestionsProp"></a>
+#### alwaysRenderSuggestions (optional)
+
+By default, suggestions are rendered only if the input field is under focus.
```
moroshko: "under focus" => "focused" (to keep the terminology consistent)

README.md
```diff
@@ -328,6 +329,13 @@ function shouldRenderSuggestions(value) {
...
+
+If you'd like to _always_ display suggestions, set `alwaysRenderSuggestions={true}`.
```
moroshko: Shall we also add a note in the `shouldRenderSuggestions` section along the lines of:
> When `shouldRenderSuggestions` returns `true`, suggestions will be rendered only when the input field is focused. If you would like to render suggestions regardless of whether the input field is focused or not, set `alwaysRenderSuggestions={true}` (`shouldRenderSuggestions` is ignored in this case).

?

test/plain-list/AutosuggestApp.js
```diff
@@ -48,6 +48,13 @@ export const shouldRenderSuggestions = sinon.spy(value => {
   return value.trim().length > 0 && value[0] !== ' ';
 });
 
+export const toggleAlwaysRenderSuggestions = sinon.spy((toggle, callback) => {
+  app.setState({
+    alwaysRenderSuggestions: toggle,
```
moroshko: "toggle" confuses me here, as I normally think about toggle as "reverse the current state". How about `setAlwaysRenderSuggestions` instead?

moroshko:
Overall looks great! Just a few minor comments above.

test/plain-list/AutosuggestApp.js
ThibWeb: Sounds good to me!

src/Autosuggest.js
ThibWeb: Erm. Feels like ESLint should have a rule for that but I haven't been able to find it.

README.md
ThibWeb: I was wondering if having the two documentation sections next to one another would be enough. Doesn't hurt to make this info more explicit.

test/plain-list/AutosuggestApp.js
ThibWeb: For those tests to be useful it is necessary to get suggestions even when there's no value in the input – that's why I use `elm` here. It's not what would happen in a real-world app but does the same job for the tests' purpose.

I added a comment to make this clearer.

ThibWeb:
As those tweaks were quite minor I just rebased them into my previous commits:

- Fix spacing in variable declaration list
- `toggleAlwaysRenderSuggestions` -> `setAlwaysRenderSuggestions`
- Documentation changes
- Comment to make the test helper's intent clearer.

src/Autosuggest.js
```diff
@@ -78,11 +79,13 @@ class Autosuggest extends Component {
...
-      const { suggestions, inputProps, shouldRenderSuggestions,
-              isCollapsed, revealSuggestions, lastAction } = nextProps;
+      const {
+        suggestions, inputProps, shouldRenderSuggestions, isCollapsed,
+        revealSuggestions, lastAction, alwaysRenderSuggestions
+      } = nextProps;
       const { value } = inputProps;
 
-      if (isCollapsed && lastAction !== 'click' && lastAction !== 'enter' &&
+      if (alwaysRenderSuggestions || isCollapsed && lastAction !== 'click' && lastAction !== 'enter' &&
```
moroshko: Do we really need `alwaysRenderSuggestions ||` here? 
All tests pass without it, so I wonder if we can just remove it, or are we missing a test here?

moroshko:
@ThibWeb Thanks for the tweaks!

I think I found an issue with the implementation.

I just added `alwaysRenderSuggestions={true}` to the CustomRenderer example (also had to change `[]` to `people` [here](https://github.com/moroshko/react-autosuggest/blob/master/demo/src/components/App/components/Examples/components/CustomRender/CustomRender.js#L18)). 
Suggestions are initially rendered, as expected. 
However, when I hit `Down`, the first suggestion isn't getting focus. Changing [this line](https://github.com/moroshko/react-autosuggest/blob/master/src/Autosuggest.js#L219) to:
```js
if (!alwaysRenderSuggestions && isCollapsed) ...
```
solves the issue, but it doesn't feel right. `isCollapsed` is `true` even though suggestions are shown.

Basically, we have a prop and a state that control the view:
* `alwaysRenderSuggestions` - prop
* `isCollapsed` - state

I feel that the prop should set the initial state. Then, `alwaysRenderSuggestions` should be used to make sure that `isCollapsed` stays `true` no matter what. This way, `isCollapsed` will be `true` if and only if suggestions are shown.

What do you think?

ThibWeb:
Yes, I've had some trouble understanding how best to interact with `isCollapsed` – doesn't surprise me that I got it wrong. I'll revisit this and probably create a standalone example, as much as a help when developing as to demonstrate what the prop is for.

bradbirnbaum:
Checking in on this pull - do we expect to take it soon?

moroshko:
@ThibWeb Have you had a chance to revisit this PR?

bradbirnbaum:
Checking in to see if we expect this to be released soon.   Thanks

ThibWeb:
Hey there – sorry for the delay, I was on holidays for a few weeks.

I still intend to work on this, and will commit to get back with a better PR _this weekend_. Hopefully this will be releasable.

jxodwyer:
@ThibWeb just wondering the status of this PR? I'd be happy to jump in and help out if you're running into issues or strapped for time.

ThibWeb:
Hey @jxodwyer, strapped for time indeed. If you want to pick this up by all means please go for it! Two weeks ago I got back onto this and made the UI example. That makes it easier to see what is going, and where it fails to work.

Useful links:

- Compiled UI example, for reference: https://rawgit.com/ThibWeb/react-autosuggest/demo/arr/demo/dist/index.html
- Branch I'm working from https://github.com/ThibWeb/react-autosuggest/tree/feature/always-render-suggestions

Next things to do IMHO:

- [ ] Look at [moroshko's suggestions/hints](https://github.com/moroshko/react-autosuggest/pull/155#issuecomment-216216991] above.
- [ ] Debug the current implementation with those hints and the examples as visual aids
- [ ] Make sure existing test suite still passes
- [ ] Add more tests for the cases that do not work here
- [ ] Update PR 🚀 
- [ ] Figure out if this example needs a Codepen link (under moroshko's account like the others?)

I don't expect this all to take a crazy amount of time (1-2 hours if you've seen the `react-autosuggest` code before).

I've added you as a contributor to https://github.com/ThibWeb/react-autosuggest, it's up to you if you prefer to use this as is or start from it in another repo 👌.

shprink:
This feature is much needed indeed! Thanks for your hard work guys, can't wait to see that merged!

ThibWeb:
@moroshko I have updated the PR, could you have a look?

I followed your [suggestions](https://github.com/moroshko/react-autosuggest/pull/155#issuecomment-216216991), `alwaysRenderSuggestions` is now used to set the default value of `isCollapsed`, and to set `shouldRenderSuggestions` to always return true. Ideally I would have wanted to completely remove any references to `alwaysRenderSuggestions` in the `Autosuggest` component, but there are a few instances where `isCollapsed` and `shouldRenderSuggestions` are not the only criteria to determine what is rendered:

- https://github.com/ThibWeb/react-autosuggest/blob/master/src/Autosuggest.js#L85
- https://github.com/ThibWeb/react-autosuggest/blob/master/src/Autosuggest.js#L185

Those extra checks are useful, but having them within the `Autosuggest` component makes the code harder to change – there are too many places in the code at different levels of abstraction that are in charge of determining whether suggestions are shown or not. Ideally I think that a single `isOpen`/`isCollapsed` prop should be the one and only source of truth for whether suggestions are shown or not. A way to get towards this would be to change how `isCollapsed` is set [`mapDispatchToProps`](https://github.com/ThibWeb/react-autosuggest/blob/master/src/Autosuggest.js#L10), but that's outside of the scope of this PR.

Other than that, I added the relevant tests to check for:

- the unwanted behavior that the previous version of this PR produced.
- the wanted behavior when selecting a suggestion

And I updated https://rawgit.com/ThibWeb/react-autosuggest/demo/arr/demo/dist/index.html to make reviewing this easier.

The last thing that's missing would be a CodePen link – not too sure how to proceed here, since this isn't something we can collaborate on via PRs?

rickhanlonii:
@moroshko just a heads up that merging this will cause some conflicts with https://github.com/moroshko/react-autosuggest/pull/165. After this is merged I'll update #165 and add any tests necessary for both props to be active at once.

ThibWeb:
Just a heads up that if someone wants to use this sooner rather than later, I have a branch where this work is compiled and npm installable: https://github.com/ThibWeb/react-autosuggest/tree/feature/arr-compiled

I'd advise _against_ relying on someone else's branch in a fork for something serious, but you can easily fork it to your own GitHub account to be more certain that it will stay how it is, and then `npm install --save react-autossugest@git://github.com/ThibWeb/react-autosuggest.git#feature/arr-compiled` with your own user / repo / branch.

moroshko:
@ThibWeb I'm keen to review and merge this PR. Do you mind resolving the merge conflicts? (sorry, they appeared after I merged #165)

carloscuatin:
@ThibWeb resolve conflicts please 😀

shprink:
🙇  @ThibWeb I will pay you a beer when you come to France if you resolve the conflicts :)

ThibWeb:
Wuhu, conflicts merged! They were quite straightforward, only had to add the `alwaysRenderSuggestions` vars where they belong.

Regarding `focusFirstSuggestion` – when both this and `alwaysRenderSuggestions` are turned on, at the moment it only puts the autosuggest focus on the first suggestion _when the user moves focus to the autosuggest field_. Focusing one of the suggestions even though the autosuggest does not have focus feels strange to me (happy to be told I'm wrong).

The running example at https://rawgit.com/ThibWeb/react-autosuggest/demo/arr/demo/dist/index.html is up to date. @moroshko this might not pass the linting since your latest config updates – I tried to make it work but `eslint-plugin-react@6.0.0-alpha.1` keeps complaining about unfound rule definitions.

LuckUY:
Hi @moroshko,

Any update on this new feature? When do you think it will be merged?

Thanks

carloscuatin:
nice @moroshko 🎉

shprink:
Sending <3 to all of you!

LuckUY:
@moroshko 

I cannot see the new "alwaysRenderSuggestions" prop on Autosuggest.js (v 3.9.0)
Did you publish a new version?

moroshko:
@LuckUY Not yet. I hope to publish new version sometime this week.

moroshko:
`alwaysRenderSuggestions` is available in [v4.0.0](https://github.com/moroshko/react-autosuggest/releases/tag/v4.0.0) now.

Issue #163, opened 2016-05-04
Add tabToEnter setting


rickhanlonii:
Certain flows allow both tab and enter to complete a suggestion (e.g. Slack user mentions). It would be great if there was a setting to allow this.

moroshko:
I must admit that I'm a bit confused with the functionality you are trying to achieve. As far as I can see, pressing Tab in Slack is somewhat equivalent to pressing Down, not pressing Enter, i.e. it moves to the next suggestion rather than selecting the focused one. The URL bar in Chrome behaves the same. [Closing this for now.](https://github.com/moroshko/react-autosuggest/pull/164#issuecomment-217735323)

PR #164, opened 2016-05-05
Tab to select

rickhanlonii:
## Overview

Adds a setting `tabToSelect` to allow tabbing to select suggestion.

Closes https://github.com/moroshko/react-autosuggest/issues/163

## Details

- Adds property `tabToSelect` (defaults to false)
- Sends `Tab` event type to `onSuggestionsUpdateRequested` and `onSuggestionSelected`
- Updates README documentation with setting and event types for above functions
- Adds tests for both true and false values of the setting

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -233,7 +233,7 @@ class Autosuggest extends Component {
             event.preventDefault();
             break;
           case 'Tab': {
-            if(tabToSelect){
+            if (tabToSelect) {
               event.preventDefault();
               const focusedSuggestion = this.getFocusedSuggestion();
 
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -51,6 +51,7 @@ class Autosuggest extends Component {
     renderSectionTitle: PropTypes.func.isRequired,
     getSectionSuggestions: PropTypes.func.isRequired,
     focusInputOnSuggestionClick: PropTypes.bool.isRequired,
+    tabToSelect: PropTypes.bool.isRequired,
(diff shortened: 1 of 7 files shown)


src/Autosuggest.js
```diff
@@ -231,7 +232,24 @@ class Autosuggest extends Component {
             }
             event.preventDefault();
             break;
-
+          case 'Tab': {
+            if (tabToSelect) {
+              event.preventDefault();
```
rickhanlonii: This may need to be moved inside the if below. Users may want to tab to another field when there's no focused suggestion.

moroshko:
@rickhanlonii Thanks a lot for a detailed PR!

I must admit that I'm a bit confused with the functionality you are trying to achieve. In #163 you mentioned "Slack user mentions". As far as I can see, pressing Tab in Slack is somewhat equivalent to pressing Down, not pressing Enter, i.e. it moves to the next suggestion rather than selecting the focused one. The URL bar in Chrome behaves the same. Am I missing something?

rickhanlonii:
Great point @moroshko, I'll update and re-open.

test/plain-list/AutosuggestApp.js
```diff
@@ -54,6 +54,18 @@ export const onSuggestionsUpdateRequested = sinon.spy(({ value }) => {
   });
 });
 
+export const tabToSelect = {
```
rickhanlonii: Note: also going to update this API per the comments [here](https://github.com/moroshko/react-autosuggest/pull/165#discussion_r62430407).

Issue #147, opened 2016-04-13
Focus first matching item


CLL80:
Is there a way to focus the first matching item? For example, when typing 'react' with the suggestions 'angular', 'react', 'react autosuggest', and 'javascript', 'react' would be focused.

moroshko:
@CLL80 At the moment, there is no API to do that. Do you know any examples online that implement this behaviour?

CLL80:
The Combobox (Dropdown + Autosize): section [here](http://alexkuz.github.io/react-input-enhancements/) does it. It would essentially be focusing the first suggestion that is returned. I tried seeing if I could pass focusedItemIndex or focusedSuggestionIndex to the AutoWhatever component from the Autosuggest component but had no luck with this.

mauriciosoares:
That's a nice feature request, I'd make use of that too...

snakeyak:
Very useful feature, would be great. Especially useful when the user types the exact value as suggested, why should he need to click on top of typing

PR #165, opened 2016-05-06
Add selectFirstSuggestion setting

rickhanlonii:
## Overview

This PR adds a setting `selectFirstSuggestion` which, when true, automatically selects the first suggestion in the list. **Note:** this PR also fixes a bug where when you hover over an option and hit enter the selected item was not completed.

Closes https://github.com/moroshko/react-autosuggest/issues/147

## Details
- Adds a `selectFirstSuggestion` prop which defaults to `false`
- Adds a method `maybeSelectFirstSuggestion` which selects the first selection when applicable
- Calls the above method in `onChange` and `onFocus`
- Adds a call to `getSuggestionValue` on Enter events to ensure the value for the selected element is input.
- Adds a call to `maybeCallOnChange` on Enter events to ensure that clients get the onChange event (this is for instances where the element was hovered or auto-selected but the client never got an onChange event).
- Adds tests for all related functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,7 @@ class Example extends React.Component {
                    onSuggestionsUpdateRequested={this.onSuggestionsUpdateRequested}
                    getSuggestionValue={getSuggestionValue}
                    renderSuggestion={renderSuggestion}
-                   inputProps={inputProps}
-                    selectFirstSuggestion={false} />
+                   inputProps={inputProps} />
     );
   }
 }
--- a/README.md
+++ b/README.md
@@ -307,6 +307,7 @@ where:
   * `'down'` - user pressed Down
   * `'up'` - user pressed Up
   * `'escape'` - user pressed Escape
+  * `'enter'` - user pressed Enter
   * `'click'` - user clicked (or tapped) on suggestion
   * `'type'` - none of the methods above (usually means that user typed something, but can also be that they pressed Backspace, pasted something into the field, etc.)
 
--- a/README.md
+++ b/README.md
@@ -309,7 +309,6 @@ where:
   * `'up'` - user pressed Up
   * `'escape'` - user pressed Escape
   * `'click'` - user clicked (or tapped) on suggestion
-  * `'auto'` - component auto-selected first suggestion
   * `'type'` - none of the methods above (usually means that user typed something, but can also be that they pressed Backspace, pasted something into the field, etc.)
 
 <a name="shouldRenderSuggestionsProp"></a>
@@ -418,7 +417,7 @@ where `isMobile` is a boolean describing whether Autosuggest operates on a mobil
 <a name="selectFirstSuggestion"></a>
 #### selectFirstSuggestion (optional)
 
-Defaulting to false, `selectFirstSuggestion={true}` will auto-select the first suggestion in the list. When selected, [onChange](#inputPropsProp) will receive the value with method `auto`.
+Defaulting to false, `selectFirstSuggestion={true}` will highlight the first suggestion in the list.
 
 ```xml
 <Autosuggest selectFirstSuggestion={true} ... />
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ class Example extends React.Component {
 * [`getSectionSuggestions`](#getSectionSuggestionsProp)
 * [`onSuggestionSelected`](#onSuggestionSelectedProp)
 * [`focusInputOnSuggestionClick`](#focusInputOnSuggestionClickProp)
-* [`selectFirstSuggestion`](#selectFirstSuggestion)
+* [`selectFirstSuggestion`](#selectFirstSuggestionProp)
 * [`theme`](#themeProp)
 * [`id`](#idProp)
 
@@ -414,7 +414,7 @@ You might want to do something like this:
 
 where `isMobile` is a boolean describing whether Autosuggest operates on a mobile device or not. You can use [kaimallea/isMobile](https://github.com/kaimallea/isMobile), for example, to determine that.
 
-<a name="selectFirstSuggestion"></a>
+<a name="selectFirstSuggestionProp"></a>
 #### selectFirstSuggestion (optional)
 
 Defaulting to false, `selectFirstSuggestion={true}` will highlight the first suggestion in the list.
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ const languages = [
 function getSuggestions(value) {
   const inputValue = value.trim().toLowerCase();
   const inputLength = inputValue.length;
-  
+
   return inputLength === 0 ? [] : languages.filter(lang =>
     lang.name.toLowerCase().slice(0, inputLength) === inputValue
   );
@@ -120,7 +120,8 @@ class Example extends React.Component {
                    onSuggestionsUpdateRequested={this.onSuggestionsUpdateRequested}
                    getSuggestionValue={getSuggestionValue}
                    renderSuggestion={renderSuggestion}
-                   inputProps={inputProps} />
+                   inputProps={inputProps}
+                    selectFirstSuggestion={false} />
     );
   }
 }
@@ -139,6 +140,7 @@ class Example extends React.Component {
 * [`getSectionSuggestions`](#getSectionSuggestionsProp)
 * [`onSuggestionSelected`](#onSuggestionSelectedProp)
 * [`focusInputOnSuggestionClick`](#focusInputOnSuggestionClickProp)
+* [`selectFirstSuggestion`](#selectFirstSuggestion)
 * [`theme`](#themeProp)
 * [`id`](#idProp)
 
@@ -307,6 +309,7 @@ where:
   * `'up'` - user pressed Up
   * `'escape'` - user pressed Escape
   * `'click'` - user clicked (or tapped) on suggestion
+  * `'auto'` - component auto-selected first suggestion
   * `'type'` - none of the methods above (usually means that user typed something, but can also be that they pressed Backspace, pasted something into the field, etc.)
 
 <a name="shouldRenderSuggestionsProp"></a>
@@ -411,6 +414,15 @@ You might want to do something like this:
 ```
 
 where `isMobile` is a boolean describing whether Autosuggest operates on a mobile device or not. You can use [kaimallea/isMobile](https://github.com/kaimallea/isMobile), for example, to determine that.
+
+<a name="selectFirstSuggestion"></a>
+#### selectFirstSuggestion (optional)
+
+Defaulting to false, `selectFirstSuggestion={true}` will auto-select the first suggestion in the list. When selected, [onChange](#inputPropsProp) will receive the value with method `auto`.
+
+```xml
+<Autosuggest selectFirstSuggestion={true} ... />
+```
 
 <a name="themeProp"></a>
 #### theme (optional)
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -157,7 +157,7 @@ class Autosuggest extends Component {
     }
   }
 
-  maybeSelectFirstSuggestion(event, value) {
+  maybeSelectFirstSuggestion() {
     const { selectFirstSuggestion, multiSection, updateFocusedSuggestion } = this.props;
 
     if (selectFirstSuggestion) {
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -83,9 +83,11 @@ class Autosuggest extends Component {
               isCollapsed, revealSuggestions, lastAction } = nextProps;
       const { value } = inputProps;
 
-      if (isCollapsed && lastAction !== 'click' && lastAction !== 'enter' &&
-          suggestions.length > 0 && shouldRenderSuggestions(value)) {
-        revealSuggestions();
+      if (suggestions.length > 0 && shouldRenderSuggestions(value)) {
+        this.maybeSelectFirstSuggestion();
+        if (isCollapsed && lastAction !== 'click' && lastAction !== 'enter') {
+          revealSuggestions();
+        }
       }
     }
   }
@@ -161,10 +163,7 @@ class Autosuggest extends Component {
     const { selectFirstSuggestion, multiSection, updateFocusedSuggestion } = this.props;
 
     if (selectFirstSuggestion) {
-      const sectionIndex = multiSection ? 0 : null;
-      const suggestionIndex = 0;
-
-      updateFocusedSuggestion(sectionIndex, suggestionIndex);
+      updateFocusedSuggestion(multiSection ? 0 : null, 0);
     }
   }
 
@@ -202,8 +201,11 @@ class Autosuggest extends Component {
       onFocus: event => {
         if (!this.justClickedOnSuggestion) {
           inputFocused(shouldRenderSuggestions(value));
-          this.maybeSelectFirstSuggestion();
           onFocus && onFocus(event);
+
+          if (suggestions.length > 0) {
+            this.maybeSelectFirstSuggestion();
+          }
         }
       },
       onBlur: event => {
@@ -225,7 +227,6 @@ class Autosuggest extends Component {
         this.maybeCallOnChange(event, value, 'type');
         inputChanged(shouldRenderSuggestions(value), 'type');
         this.maybeCallOnSuggestionsUpdateRequested({ value, reason: 'type' });
-        this.maybeSelectFirstSuggestion();
       },
       onKeyDown: (event, data) => {
         switch (event.key) {
@@ -260,7 +261,7 @@ class Autosuggest extends Component {
 
             closeSuggestions('enter');
 
-            if (focusedSuggestion) {
+            if (focusedSuggestion !== null) {
               const newValue = getSuggestionValue(focusedSuggestion);
 
               onSuggestionSelected(event, {
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -243,7 +243,7 @@ class Autosuggest extends Component {
                 // valueBeforeUpDown can be null if, for example, user
                 //  hovers on the first suggestion and then pressed Up.
                 //  if that happens, use the original input value
-                newValue = valueBeforeUpDown ? valueBeforeUpDown : value;
+                newValue = (valueBeforeUpDown === null ? value : valueBeforeUpDown);
               } else {
                 newValue = this.getSuggestionValueByIndex(newFocusedSectionIndex, newFocusedItemIndex);
               }
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -237,10 +237,16 @@ class Autosuggest extends Component {
               }
             } else if (suggestions.length > 0) {
               const { newFocusedSectionIndex, newFocusedItemIndex } = data;
-              const safeValue = valueBeforeUpDown ? valueBeforeUpDown : value;
-              const newValue = newFocusedItemIndex === null ?
-                safeValue :
-                this.getSuggestionValueByIndex(newFocusedSectionIndex, newFocusedItemIndex);
+
+              let newValue;
+              if(newFocusedItemIndex === null){
+                // valueBeforeUpDown can be null if, for example, user
+                //  hovers on the first suggestion and then pressed Up.
+                //  if that happens, use the original input value
+                newValue = valueBeforeUpDown ? valueBeforeUpDown : value;
+              } else {
+                newValue = this.getSuggestionValueByIndex(newFocusedSectionIndex, newFocusedItemIndex);
+              }
 
               updateFocusedSuggestion(newFocusedSectionIndex, newFocusedItemIndex, value);
               this.maybeCallOnChange(event, newValue, event.key === 'ArrowDown' ? 'down' : 'up');
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -164,7 +164,7 @@ class Autosuggest extends Component {
       let sectionIndex = multiSection ? 0 : null;
       let suggestionIndex = 0;
 
-      updateFocusedSuggestion(sectionIndex, suggestionIndex, value);
+      updateFocusedSuggestion(sectionIndex, suggestionIndex);
       this.maybeCallOnChange(event, value, 'auto');
     }
   }
@@ -192,7 +192,8 @@ class Autosuggest extends Component {
       getSectionSuggestions, focusInputOnSuggestionClick, theme, isFocused,
       isCollapsed, focusedSectionIndex, focusedSuggestionIndex,
       valueBeforeUpDown, inputFocused, inputBlurred, inputChanged,
-      updateFocusedSuggestion, revealSuggestions, closeSuggestions
+      updateFocusedSuggestion, revealSuggestions, closeSuggestions,
+      getSuggestionValue
     } = this.props;
     const { value, onBlur, onFocus, onKeyDown } = inputProps;
     const isOpen = isFocused && !isCollapsed && this.willRenderSuggestions();
@@ -223,7 +224,7 @@ class Autosuggest extends Component {
         const { shouldRenderSuggestions } = this.props;
 
         this.maybeCallOnChange(event, value, 'type');
-        inputChanged(shouldRenderSuggestions(value), 'type', suggestions);
+        inputChanged(shouldRenderSuggestions(value), 'type');
         this.maybeCallOnSuggestionsUpdateRequested({ value, reason: 'type' });
         this.maybeSelectFirstSuggestion(event, value);
       },
@@ -237,8 +238,9 @@ class Autosuggest extends Component {
               }
             } else if (suggestions.length > 0) {
               const { newFocusedSectionIndex, newFocusedItemIndex } = data;
+              const safeValue = valueBeforeUpDown ? valueBeforeUpDown : value;
               const newValue = newFocusedItemIndex === null ?
-                valueBeforeUpDown :
+                safeValue :
                 this.getSuggestionValueByIndex(newFocusedSectionIndex, newFocusedItemIndex);
 
               updateFocusedSuggestion(newFocusedSectionIndex, newFocusedItemIndex, value);
@@ -251,14 +253,18 @@ class Autosuggest extends Component {
             const focusedSuggestion = this.getFocusedSuggestion();
 
             if (focusedSuggestion !== null) {
+              const newValue = getSuggestionValue(focusedSuggestion);
+
               closeSuggestions('enter');
               onSuggestionSelected(event, {
                 suggestion: focusedSuggestion,
-                suggestionValue: value,
+                suggestionValue: newValue,
                 sectionIndex: focusedSectionIndex,
                 method: 'enter'
               });
-              this.maybeCallOnSuggestionsUpdateRequested({ value, reason: 'enter' });
+
+              this.maybeCallOnChange(event, newValue, event.key === 'enter');
+              this.maybeCallOnSuggestionsUpdateRequested({ value: newValue, reason: 'enter' });
             }
             break;
           }
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -51,6 +51,7 @@ class Autosuggest extends Component {
     renderSectionTitle: PropTypes.func.isRequired,
     getSectionSuggestions: PropTypes.func.isRequired,
     focusInputOnSuggestionClick: PropTypes.bool.isRequired,
+    selectFirstSuggestion: PropTypes.bool.isRequired,
     theme: PropTypes.object.isRequired,
     id: PropTypes.string.isRequired,
     inputRef: PropTypes.func.isRequired,
@@ -156,6 +157,18 @@ class Autosuggest extends Component {
     }
   }
 
+  maybeSelectFirstSuggestion(event, value) {
+    const { selectFirstSuggestion, multiSection, updateFocusedSuggestion } = this.props;
+
+    if (selectFirstSuggestion) {
+      let sectionIndex = multiSection ? 0 : null;
+      let suggestionIndex = 0;
+
+      updateFocusedSuggestion(sectionIndex, suggestionIndex, value);
+      this.maybeCallOnChange(event, value, 'auto');
+    }
+  }
+
   willRenderSuggestions() {
     const { suggestions, inputProps, shouldRenderSuggestions } = this.props;
     const { value } = inputProps;
@@ -189,6 +202,7 @@ class Autosuggest extends Component {
       onFocus: event => {
         if (!this.justClickedOnSuggestion) {
           inputFocused(shouldRenderSuggestions(value));
+          this.maybeSelectFirstSuggestion(event, value);
           onFocus && onFocus(event);
         }
       },
@@ -209,8 +223,9 @@ class Autosuggest extends Component {
         const { shouldRenderSuggestions } = this.props;
 
         this.maybeCallOnChange(event, value, 'type');
-        inputChanged(shouldRenderSuggestions(value), 'type');
+        inputChanged(shouldRenderSuggestions(value), 'type', suggestions);
         this.maybeCallOnSuggestionsUpdateRequested({ value, reason: 'type' });
+        this.maybeSelectFirstSuggestion(event, value);
       },
       onKeyDown: (event, data) => {
         switch (event.key) {
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -165,7 +165,6 @@ class Autosuggest extends Component {
       let suggestionIndex = 0;
 
       updateFocusedSuggestion(sectionIndex, suggestionIndex);
-      this.maybeCallOnChange(event, value, 'auto');
     }
   }
 
@@ -252,7 +251,7 @@ class Autosuggest extends Component {
           case 'Enter': {
             const focusedSuggestion = this.getFocusedSuggestion();
 
-            if (focusedSuggestion) {
+            if (focusedSuggestion !== null) {
               const newValue = getSuggestionValue(focusedSuggestion);
 
               closeSuggestions('enter');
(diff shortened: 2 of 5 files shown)


rickhanlonii:
@moroshko there will be some serious merge conflicts with this branch and https://github.com/moroshko/react-autosuggest/pull/164. Let me know when you merge one or the other and I can handle the conflicts.

src/Autosuggest.js
```diff
@@ -156,6 +157,18 @@ class Autosuggest extends Component {
...
 
+  maybeSelectFirstSuggestion(event, value) {
+    const { selectFirstSuggestion, multiSection, updateFocusedSuggestion } = this.props;
+
+    if (selectFirstSuggestion) {
+      let sectionIndex = multiSection ? 0 : null;
+      let suggestionIndex = 0;
+
+      updateFocusedSuggestion(sectionIndex, suggestionIndex);
+      this.maybeCallOnChange(event, value, 'auto');
```
moroshko: Could you explain why do we call `maybeCallOnChange()` here?

src/Autosuggest.js
```diff
@@ -222,8 +238,9 @@ class Autosuggest extends Component {
               }
             } else if (suggestions.length > 0) {
               const { newFocusedSectionIndex, newFocusedItemIndex } = data;
+              const safeValue = valueBeforeUpDown ? valueBeforeUpDown : value;
               const newValue = newFocusedItemIndex === null ?
-                valueBeforeUpDown :
+                safeValue :
```
moroshko: Could you explain this change please?

src/Autosuggest.js
```diff
@@ -235,15 +252,19 @@ class Autosuggest extends Component {
           case 'Enter': {
             const focusedSuggestion = this.getFocusedSuggestion();
 
-            if (focusedSuggestion !== null) {
+            if (focusedSuggestion) {
```
moroshko: I prefer the explicit `focusedSuggestion !== null`. Do you mind if we keep it?

src/Autosuggest.js
```diff
@@ -235,15 +252,19 @@ class Autosuggest extends Component {
...
               onSuggestionSelected(event, {
                 suggestion: focusedSuggestion,
-                suggestionValue: value,
+                suggestionValue: newValue,
                 sectionIndex: focusedSectionIndex,
                 method: 'enter'
               });
-              this.maybeCallOnSuggestionsUpdateRequested({ value, reason: 'enter' });
+
+              this.maybeCallOnChange(event, newValue, event.key === 'enter');
```
moroshko: Shouldn't the last parameter be just `'enter'`?

src/AutosuggestContainer.js
```diff
@@ -69,13 +69,14 @@ export default class AutosuggestContainer extends Component {
...
     focusInputOnSuggestionClick: PropTypes.bool,
+    selectFirstSuggestion: PropTypes.bool,
     theme: PropTypes.object,
     id: PropTypes.string
   };
 
   static defaultProps = {
     onSuggestionsUpdateRequested: noop,
-    shouldRenderSuggestions: value => value.trim().length > 0,
+    shouldRenderSuggestions: value => value && value.trim().length > 0,
```
moroshko: Why is this change?

test/plain-list/AutosuggestApp.test.js
```diff
@@ -131,6 +132,83 @@ describe('Plain list Autosuggest', () => {
     });
   });
 
+  describe('when typing and matches exist (when selectFirstSuggestion is true)', () => {
+    before(() => {
+      selectFirstSuggestion.true();
+    });
+
+    after(() => {
+      selectFirstSuggestion.false();
```
moroshko: I really don't like this API. How about:
```js
setSelectFirstSuggestion(true); // or false
```
?

src/Autosuggest.js
rickhanlonii: This is intentionally changed because it's possible for `focusedSuggestion` to be `undefined` instead of `null`. We could change it to: 

`focusedSuggestion !== null && typeof(focusedSuggestion) !== 'undefined'`

but since this is always an object reference there's no gain there over the implicit check. Totally your call on the semantics but we do need to handle the `undefined` case here.

src/Autosuggest.js
rickhanlonii: Yes, good catch.

src/AutosuggestContainer.js
rickhanlonii: It's possible for the value to be null (either from default props in the library, or from clients setting the value to null). And if the value is null or undefined, there's no suggestions to render. Does that make sense?

I did just realize that this will return false if value is 0, so I'll need to update this to be more explicit.

test/plain-list/AutosuggestApp.test.js
rickhanlonii: Yeah I wasn't crazy about it either. I was trying to mirror the semantics of the `spy.reset()` calls. I'll update it.

rickhanlonii:
Thanks for the review @moroshko, I'll make the updates and re-open.

src/Autosuggest.js
rickhanlonii: To mirror the call to `maybeCallOnChange` in the up/town events. I.e. the system is auto-selecting the first item, which is a change event, and clients will want to handle the selection change appropriately. Let me know if that makes sense.

src/Autosuggest.js
rickhanlonii: Sure thing. 

The `valueBeforeUpDown` variable isn't always set and can be null, which raises an exception later in the chain . 

This is because when we call `updateFocusedSuggestion` in the mouse events we don't pass a `value` (and can't since there's no reference to it). So if you mouse over an item before up/down, `valueBeforeUpDown` stays null. If you then transition directly to the input field using up/down, then `valueBeforeUpDown` is still null but `value` is equal to the input field.

To reproduce:
- type 'c'
- mouse over the first suggestion
- press Up
- you will get a console error
- this also applies to the last suggestion and pressing down

This check ensures that if there's no `valueBeforeUpDown`, then `value`is used (which can only happen when you transition into the input field, and thus `value` is the value of the input field).

Here's a GIF of the bug (note: this is live on master, so this is unrelated to my changes):

src/Autosuggest.js
rickhanlonii: Thinking about this a little more, the auto-select should perform more like a hover than an updown. You probably wouldn't want the input value to change as a soon as it auto-selects the first item--you could never type anything without the input changing out from under you. I think it's still useful to have as it's own type `auto`, but let me know what you think.

moroshko: I think we should trigger `onChange` only when the input value changes, which is not the case here (I agree, it's more similar to "on hover" behaviour, which doesn't change the input value). So, I suggest we don't call `this.maybeCallOnChange()` here.

src/Autosuggest.js
moroshko: Great, thanks for the detailed explanation and for fixing this bug!
Can I just suggest that we make the code a little more self explanatory:
```js
const safeValue = valueBeforeUpDown === null // This can happen if, for example, user hovers on the first suggestion and then clicks Up
  ? value 
  : valueBeforeUpDown;
```

src/Autosuggest.js
moroshko: Could you describe how to get to the state when `focusedSuggestion === undefined`?
Did the fact that we didn't check for `undefined` caused any issues?
Do we have a missing test case?

src/Autosuggest.js
moroshko: We should have a test case that would catch this!

src/Autosuggest.js
moroshko: Let's also add a test case for this scenario.

src/AutosuggestContainer.js
moroshko: Default props in the library? Where is that?
If the consumer doesn't set `value` to a string, I'd prefer to throw an error in this case. `value` should always be a string.

moroshko:
@rickhanlonii By the way, why did you close the PR? Leave it open, and force push your changes.

src/AutosuggestContainer.js
rickhanlonii: > Default props in the library? Where is that?

`valueBeforeUpDown` [defaults to null](https://github.com/rickhanlonii/react-autosuggest/blob/e5336c6e890db37ceb5cd2a1f9eb60d33c2f210a/src/AutosuggestContainer.js#L102-L102) and is set to null in [a few places](https://github.com/rickhanlonii/react-autosuggest/blob/fde093db1a4630661f31cf1ae4d5f309b2c39de9/src/reducerAndActions.js#L106), which means that clients can receive `null` in `onChange` events. If they don't check for null and convert to the empty string (see the demos, which just set the `value` to what's passed), they can set `value` to null. 

This is what happens in https://github.com/moroshko/react-autosuggest/pull/165#discussion_r62436123.

So the issue here isn't that clients are passing `null` it's that they're passing through a `null` given to them by the library.

src/Autosuggest.js
rickhanlonii: Sure can. 

This is something new that has been introduced and wasn't possible before (but as I'll explain, is unavoidable as far as I can tell).

 Theoretically, it happens when a selection is hovered but that index is not in the suggestions. Of course, this can't happen on hover events because the suggestions are not visible (and don't exist). But `selectFirstSuggestion` acts as a hover event and selects/hovers the first suggestion `onChange`. Since `onChange` has no way to know if there are suggestions (suggestions don't exist until the next render cycle) it always sets the `focusedSuggestionIndex` to 0. This is the part that I believe is unavoidable.

So calling `this.getFocusedSuggestion()` will return `undefined` at the focused index, every time there are no suggestions given and enter is pressed:



Of course a better solution here is to not set a `focusedSuggestionIndex` if the `suggestions` array is empty. I looked at this pretty extensively, and I found no way to determine this at the time `onChange` is called (since the suggestions are not fetched until the next render cycle).

Thus I think a fair strategy is to say: when enter is pressed and the focused selection doesn't exist in the selections array, do nothing.

Does that make sense?

rickhanlonii:
@moroshko I typically close PRs if they need a non-trivial revision to keep open PRs clean.

src/Autosuggest.js
moroshko: @rickhanlonii Are we going to delete L168?

src/AutosuggestContainer.js
moroshko: Makes sense! How about we make sure that we never pass `null` to `shouldRenderSuggestions()` instead?
You fixed the problem for those who don't define their own `shouldRenderSuggestions()`, but those who do define their own will still get `null`, right?

moroshko:
@rickhanlonii Let me know when you update the PR, and I'll review it again.

src/Autosuggest.js
rickhanlonii: @moroshko sorry, I didn't ignore you, I added the +1 reaction to agree.

src/Autosuggest.js
rickhanlonii: @moroshko let me know if this makes sense.

src/AutosuggestContainer.js
rickhanlonii: Yeah, actually I think this is unnecessary with the fix in https://github.com/moroshko/react-autosuggest/pull/165#discussion_r62436123 defaulting to the input value.

src/Autosuggest.js
rickhanlonii: Tests added here: https://github.com/moroshko/react-autosuggest/pull/165/commits/4764c0496d8fb6c0f6ceddef12d5c548a50be5a1

src/Autosuggest.js
rickhanlonii: @moroshko I added a test for this. If you keep this as `!= null` this test till fail: 

https://github.com/moroshko/react-autosuggest/pull/165/commits/36bb114effa5d2ef6a4b1ed96f018138300c1440#diff-d6150ad17107b5367bf67fd9cae74c09R257

src/Autosuggest.js
rickhanlonii: Tests added here: https://github.com/moroshko/react-autosuggest/pull/165/commits/c27e4156deacdcbac87a651d35a9b29431ee2eed

src/AutosuggestContainer.js
rickhanlonii: Ok, I removed this and the test added for https://github.com/moroshko/react-autosuggest/pull/165#discussion_r62436123 confirm that the `safeValue` change prevents this value from being null.

test/plain-list/AutosuggestApp.test.js
rickhanlonii: Updated in https://github.com/moroshko/react-autosuggest/pull/165/commits/ab3ba610685b26793f328af13e0818ebe0a4b148

rickhanlonii:
@moroshko, updated and ready for review.

src/Autosuggest.js
```diff
@@ -156,6 +157,17 @@ class Autosuggest extends Component {
     }
   }
 
+  maybeSelectFirstSuggestion(event, value) {
```
moroshko: `event` and `value` are not used anymore, so we should remove them.

src/Autosuggest.js
```diff
@@ -156,6 +157,17 @@ class Autosuggest extends Component {
...
+    const { selectFirstSuggestion, multiSection, updateFocusedSuggestion } = this.props;
+
+    if (selectFirstSuggestion) {
+      let sectionIndex = multiSection ? 0 : null;
+      let suggestionIndex = 0;
```
moroshko: Can we make them `const`?

README.md
```diff
@@ -139,6 +140,7 @@ class Example extends React.Component {
 * [`getSectionSuggestions`](#getSectionSuggestionsProp)
 * [`onSuggestionSelected`](#onSuggestionSelectedProp)
 * [`focusInputOnSuggestionClick`](#focusInputOnSuggestionClickProp)
+* [`selectFirstSuggestion`](#selectFirstSuggestion)
```
moroshko: The convention in this project is to have `Prop` at the end, so `selectFirstSuggestion` => `selectFirstSuggestionProp`

README.md
```diff
@@ -120,7 +120,8 @@ class Example extends React.Component {
                    onSuggestionsUpdateRequested={this.onSuggestionsUpdateRequested}
                    getSuggestionValue={getSuggestionValue}
                    renderSuggestion={renderSuggestion}
-                   inputProps={inputProps} />
+                   inputProps={inputProps}
+                    selectFirstSuggestion={false} />
```
moroshko: Extra space?

src/Autosuggest.js
```diff
@@ -189,6 +202,7 @@ class Autosuggest extends Component {
       onFocus: event => {
         if (!this.justClickedOnSuggestion) {
           inputFocused(shouldRenderSuggestions(value));
+          this.maybeSelectFirstSuggestion(event, value);
```
moroshko: No need to pass params anymore

src/Autosuggest.js
```diff
@@ -211,6 +225,7 @@ class Autosuggest extends Component {
         this.maybeCallOnChange(event, value, 'type');
         inputChanged(shouldRenderSuggestions(value), 'type');
         this.maybeCallOnSuggestionsUpdateRequested({ value, reason: 'type' });
+        this.maybeSelectFirstSuggestion(event, value);
```
moroshko: Same. No need to pass params anymore.

src/Autosuggest.js
```diff
@@ -222,9 +237,16 @@ class Autosuggest extends Component {
...
-              const newValue = newFocusedItemIndex === null ?
-                valueBeforeUpDown :
-                this.getSuggestionValueByIndex(newFocusedSectionIndex, newFocusedItemIndex);
+
+              let newValue;
+              if (newFocusedItemIndex === null) {
+                // valueBeforeUpDown can be null if, for example, user
+                //  hovers on the first suggestion and then pressed Up.
+                //  if that happens, use the original input value
+                newValue = valueBeforeUpDown ? valueBeforeUpDown : value;
```
moroshko: Let's be as explicit as we can here:
```js
newValue = (valueBeforeUpDown === null ? value : valueBeforeUpDown);
```

test/plain-list/AutosuggestApp.js
```diff
@@ -54,6 +54,11 @@ export const onSuggestionsUpdateRequested = sinon.spy(({ value }) => {
   });
 });
 
+let selectFirstSuggestionValue = false;
```
moroshko: Let's call it simply `selectFirstSuggestion`, if you don't object?

test/plain-list/AutosuggestApp.js
```diff
@@ -54,6 +54,11 @@ export const onSuggestionsUpdateRequested = sinon.spy(({ value }) => {
...
+export function setSelectFirstSuggestion(value){
```
moroshko: Did you run `npm run lint`? I would expect ESLint to complain that a space is missing before `{`.

src/Autosuggest.js
moroshko: @rickhanlonii Yeah, I see :)

What I don't feel comfortable about is that `UPDATE_FOCUSED_SUGGESTION` is triggered too much, even when we don't update the focused suggestion.

What do you think about calling `maybeSelectFirstSuggestion()` in `componentWillReceiveProps()` instead?
Have a look [here](https://github.com/moroshko/react-autosuggest/commit/c2cc05484a0b8a459f65f115177aa8be5dec7d90).

It is more code, but I think it's cleaner.
What do you think?

cpsubrian:
Excited to make use of this for my tagging UI! Please let me know if I can help.

ethanve:
Any news on this PR. Would love to use this feature!

rickhanlonii:
I'll get the few updates in, sorry it's taking so long.

README.md
rickhanlonii: I think a lot of people go straight to the example code first so it's valuable to have properties listed there with the default values so that they're at least aware they exist. This is handled similarly in `redux-devtools` [here](https://github.com/gaearon/redux-devtools/commit/7f3cdf3b48b789984549bd862b41707b956f247a). Maybe there could/should be a comment.

Thoughts?

src/Autosuggest.js
rickhanlonii: Looks great! I moved the changes on that branch into this one.

rickhanlonii: Update: After incorporating the changes [here](https://github.com/moroshko/react-autosuggest/commit/c2cc05484a0b8a459f65f115177aa8be5dec7d90) we can explicitly check for null since `maybeSelectFirstSuggestion` isn't called for every `onChange` anymore.

rickhanlonii:
@moroshko I believe this is all set, please review.

README.md
moroshko: The example in `redux-devtools` is much smaller, so adding one more prop there makes sense. Here, I feel like the Basic Usage example is already too big, which can be a barrier for some people. Personally, if I wanted to highlight the first suggestion and couldn't immediately see a prop for it, I'd click Cmd+F and type "first", which will reveal the desired prop.

src/Autosuggest.js
```diff
@@ -156,6 +159,14 @@ class Autosuggest extends Component {
     }
   }
 
+  maybeSelectFirstSuggestion() {
+    const { selectFirstSuggestion, multiSection, updateFocusedSuggestion } = this.props;
+
+    if (selectFirstSuggestion) {
+      updateFocusedSuggestion(multiSection ? 0 : null, 0);
+    }
+  }
```
moroshko: When `selectFirstSuggestion` is `false`, this function will be called over and over again just to perform the destructuring and the `if` check. If wonder whether it worth to set `maybeSelectFirstSuggestion` to `noop` in the constructor if we know that `selectFirstSuggestion === false` (this is obviously assuming that `selectFirstSuggestion` cannot be changed during the lifetime of an Autosuggest). Do you think it worth the effort?

src/Autosuggest.js
```diff
@@ -238,13 +262,17 @@ class Autosuggest extends Component {
...
               onSuggestionSelected(event, {
                 suggestion: focusedSuggestion,
-                suggestionValue: value,
+                suggestionValue: newValue,
                 sectionIndex: focusedSectionIndex,
                 method: 'enter'
               });
-              this.maybeCallOnSuggestionsUpdateRequested({ value, reason: 'enter' });
+
+              this.maybeCallOnChange(event, newValue, 'enter');
```
moroshko: Let's add `'enter'` to the possible `method` values [in the docs](https://github.com/moroshko/react-autosuggest#inputprops-required):

    'enter' - user pressed Enter after hovering with mouse over a suggestion

test/plain-list/AutosuggestApp.test.js
```diff
@@ -131,6 +132,102 @@ describe('Plain list Autosuggest', () => {
...
+    after(() => {
+      setSelectFirstSuggestion(false);
+    });
+
+    beforeEach(() => {
+      onSuggestionSelected.reset();
+      focusAndSetInputValue('p');
+    });
+
+    it('should focus on any suggestion', () => {
```
moroshko: any => the first

test/plain-list/AutosuggestApp.test.js
```diff
@@ -131,6 +132,102 @@ describe('Plain list Autosuggest', () => {
...
+      expectFocusedSuggestion('Perl');
+    });
+
+    it('should hide suggestions when Escape is pressed', () => {
+      clickEscape();
+      expectSuggestions([]);
+    });
```
moroshko: This is one example of a repeated test (same test for `selectFirstSuggestion = true` and `selectFirstSuggestion = false`). Although I can see the value in this repetition, I'm not sure it worth the overhead it adds to the maintenance of the test suite. Just by adding this `selectFirstSuggestion` boolean prop to Autosuggest, this test file grew from 665 to 1025 lines! I suggest we add only tests that are specific to `selectFirstSuggestion = true`. Thoughts?

test/plain-list/AutosuggestApp.test.js
```diff
@@ -131,6 +132,102 @@ describe('Plain list Autosuggest', () => {
...
+      expectInputValue('');
+    });
+
+    it('should show suggestions when input is focused again', () => {
+      blurInput();
+      focusInput();
+      expectSuggestions(['Perl', 'PHP', 'Python']);
+    });
+
+    it('should focus suggestion when input is focused again', () => {
```
moroshko: suggestion => on the first suggestion

test/plain-list/AutosuggestApp.test.js
```diff
@@ -131,6 +132,102 @@ describe('Plain list Autosuggest', () => {
...
+        suggestionValue: 'Perl',
+        sectionIndex: null,
+        method: 'enter'
+      });
+
+      expect(onChange).to.have.been.calledOnce;
+      expect(onChange).to.be.calledWith(eventInstance, {
+        newValue: 'Perl',
+        method: 'enter'
+      });
```
moroshko: There is a special section `describe('inputProps.onChange', ...)`. Let's move this test closer to it.

test/plain-list/AutosuggestApp.test.js
```diff
@@ -131,6 +132,102 @@ describe('Plain list Autosuggest', () => {
...
+    it('should be called once with the right parameters when Enter is pressed', () => {
+      onChange.reset();
+      clickEnter();
+      expect(onSuggestionSelected).to.have.been.calledOnce;
+      expect(onSuggestionSelected).to.have.been.calledWithExactly(eventInstance, {
+        suggestion: { name: 'Perl', year: 1987 },
+        suggestionValue: 'Perl',
+        sectionIndex: null,
+        method: 'enter'
+      });
```
moroshko: There is a special section `describe('onSuggestionSelected', ...)`. Let's move this test closer to it.

test/plain-list/AutosuggestApp.test.js
```diff
@@ -140,13 +237,84 @@ describe('Plain list Autosuggest', () => {
...
+    beforeEach(() => {
+      focusAndSetInputValue('z');
+    });
+
+    it('should not show suggestions', () => {
+      expectSuggestions([]);
+    });
+
+    it('should not focus suggestions', () => {
+      expectFocusedSuggestion(null);
```
moroshko: What's the value of this test given we know from the test above that no suggestions are rendered?

test/plain-list/AutosuggestApp.test.js
```diff
@@ -140,13 +237,84 @@ describe('Plain list Autosuggest', () => {
...
+    });
+
+    describe('when pressing up', () => {
+      beforeEach(() => {
+        clickUp();
+      });
+
+      it('should show the original input value', () => {
+        expectInputValue('p');
+      });
```
moroshko: Thanks for fixing this bug and adding a test!

test/plain-list/AutosuggestApp.test.js
```diff
@@ -140,13 +237,84 @@ describe('Plain list Autosuggest', () => {
...
+    });
   });
 
   describe('when pressing Down', () => {
+
```
moroshko: Can we delete this new line?

test/plain-list/AutosuggestApp.test.js
```diff
@@ -140,13 +237,84 @@ describe('Plain list Autosuggest', () => {
...
+      expectFocusedSuggestion(null);
+    });
+
+    it('should clear the input when Escape is pressed', () => {
+      clickEscape();
+      expectInputValue('');
+    });
+  });
+
+  describe('when typing and matches do not exist (when selectFirstSuggestion is true)', () => {
```
moroshko: Again, there is a lot of repetition between this `describe` and the one above. Let's add only tests that are specific to the new feature (`selectFirstSuggestion = true`).

test/plain-list/AutosuggestApp.test.js
```diff
@@ -229,6 +479,25 @@ describe('Plain list Autosuggest', () => {
...
+    });
+
+    beforeEach(() => {
+      focusAndSetInputValue('p');
+    });
+
+    it('should hide suggestions since the first suggestion is focused', () => {
+      clickEnter();
+      expectSuggestions([]);
+    });
```
moroshko: Do you really feel like this test adds value given we have [this test](https://github.com/moroshko/react-autosuggest/blob/master/test/plain-list/AutosuggestApp.test.js#L220-L224)? I personally feel it is unnecessary.

test/plain-list/AutosuggestApp.test.js
```diff
@@ -244,6 +513,29 @@ describe('Plain list Autosuggest', () => {
...
+      expectInputValue('');
+    });
+
+    it('should reset the input if suggestions are hidden but were shown before', () => {
+      focusAndSetInputValue('p');
+      focusAndSetInputValue('pz');
+      clickEscape();
+      expectInputValue('');
+    });
+  });
```
moroshko: Same here. It's just a copy of the `describe` above.

test/plain-list/AutosuggestApp.test.js
```diff
@@ -284,6 +576,38 @@ describe('Plain list Autosuggest', () => {
...
+      expect(getSuggestionValue).to.have.been.calledOnce;
+      expect(getSuggestionValue).to.have.been.calledWithExactly({ name: 'PHP', year: 1995 });
+    });
+
+    it('should be called once with the right parameters when suggestion is clicked', () => {
+      clickSuggestion(2);
+      expect(getSuggestionValue).to.have.been.calledOnce;
+      expect(getSuggestionValue).to.have.been.calledWithExactly({ name: 'Python', year: 1991 });
+    });
+  });
```
moroshko: I suggest instead of adding this `describe` we add one more test to the `describe` above:

    it('should be called once with the right parameters when input is focused', ...)

test/plain-list/AutosuggestApp.test.js
```diff
@@ -406,6 +730,29 @@ describe('Plain list Autosuggest', () => {
...
+
+    it('should be called with the right parameters when user types', () => {
+      focusAndSetInputValue('c');
+      expect(onChange).to.have.been.calledOnce;
+      expect(onChange).to.be.calledWith(eventInstance, {
+        newValue: 'c',
+        method: 'type'
+      });
+    });
+  });
```
moroshko: How is this test different from [this one](https://github.com/moroshko/react-autosuggest/blob/master/test/plain-list/AutosuggestApp.test.js#L317-L324)?

moroshko: `onChange` is called now with `method='enter'`. Can we add a test to cover this?

test/plain-list/AutosuggestApp.test.js
```diff
@@ -456,6 +803,19 @@ describe('Plain list Autosuggest', () => {
...
+      clickDown();
+      clickEnter();
+
+      // onChange should not have been called because the value didn't change
+      expect(onChange).to.have.been.calledExactlyOnce;
+      expect(onChange).to.be.calledWithExactly(eventInstance, {
+        newValue: 'Java',
+        method: 'down'
+      });
+    });
```
moroshko: 1. The description says "should not call onChange", but we expect it to be called once. What am I missing here?
2. If we exercise `onChange` behaviour, let's put the test in the `describe('inputProps.onChange', ...)` section

moroshko:
@rickhanlonii Review done! (sorry it took me so long)

src/Autosuggest.js
moroshko: Thinking more about this, I feel like it doesn't worth the effort. Interested to hear what others think.

rickhanlonii:
Hey @moroshko, thanks for the review! I'll review those comments and make changes.

src/Autosuggest.js
rickhanlonii: I agree that would be an improvement, but it's probably better to change it post-merge. There's a lot invested in this PR now and it would be wise not to introduce more bugs and testing so late in the game.

src/Autosuggest.js
rickhanlonii: I'm going to leave this as `user pressed Enter` as it will happen any time enter is pressed and the values are different. Using the suggested language here neglects to mention the case where you press enter when the first suggestion is auto-selected, and would require updating for any future cases with an increasingly long explanation.

test/plain-list/AutosuggestApp.test.js
rickhanlonii: I think there's a line here and it's your call, so the best strategy here may be for you to merge the PR and remove any tests you think are unnecessary. 

Fwiw though, I went back and forth on this and opted to add all the tests once I noticed that there were actually differences that popped up and needed fixed by having them.

Another option may also be to move the `selectFirstSuggestion = true` tests to a separate file. But as I said, do whatever you think is best.

test/plain-list/AutosuggestApp.test.js
rickhanlonii: Admittedly, right now `expectFocusedSuggestion` checks for the existence of DOM elements but that may be updated to include checking the focused index in the component to be more thorough.

test/plain-list/AutosuggestApp.test.js
rickhanlonii: This test was actually added during discussion [here](https://github.com/moroshko/react-autosuggest/pull/165/commits/36bb114effa5d2ef6a4b1ed96f018138300c1440#diff-d6150ad17107b5367bf67fd9cae74c09R257). So yes, this one is actually necessary. 

What's important to note is that there's different internal logic when pressing up/down to select a suggestion and when auto-selecting or mousing over a suggestion. So it's not correct to assume that tests using keyboard inputs are valid/repetitive tests for mouse inputs or system actions. They're truly handled differently and need to be tested independently.

test/plain-list/AutosuggestApp.test.js
rickhanlonii: Here `setSelectFirstSuggestion` is set to true. This is testing that `onChange` is only called once -- i.e. that `onChange` was not called when auto-selecting the first suggestion.

test/plain-list/AutosuggestApp.test.js
rickhanlonii: That should say `should not call onChange for the 'Enter' event` to be more clear, since `onChange` will be called, and called only for, the `Down` event.   
  
I'll update the wording and move.

rickhanlonii: Ok, I updated the logic here to `onChange.clear` after the Down event, so it should be more clear that we're testing to make sure there's no `onChange` for only the `Enter` event.

See https://github.com/moroshko/react-autosuggest/pull/165/commits/19e5409f5f05cc5f0f633d967ef3c544dfec22cb

test/plain-list/AutosuggestApp.test.js
rickhanlonii: This did remind me that we needed the same test for mouse hover events, which I added in https://github.com/moroshko/react-autosuggest/pull/165/commits/a3dba9f3e5e3d2e812fe9bb84abe997b135ddfc5

rickhanlonii:
He @moroshko, I think I made all the updated you requested. I think with respect to the tests you should exercise ownership freedom and remove/change whatever you prefer.

Let me know if you have any questions!

moroshko:
@rickhanlonii Thanks a lot for all your efforts!! I'll take it from here.

moroshko:
I ended up renaming `selectFirstSuggestion` to `focusFirstSuggestion` as in the current terminology "selecting a suggestion" means clicking it (or pressing Enter).

moroshko:
[v3.8.0](https://github.com/moroshko/react-autosuggest/releases/tag/v3.8.0) is on npm now. Thanks a lot for this contribution!

rickhanlonii:
Great, thanks!

PR #170, opened 2016-05-17
Blur input on touch devices when clicking on the rest of the page.

andysprout:
### Before this change

On Safari (Mac OS), Chrome (Mac OS and android and Mac OS with touch device emulation):
- when the user clicks on the `<input>` element
  - then the `<input>` element is focused
- when the user clicks or taps outside the react-autosuggest component
  - then the `<input>` element loses focus

On Mobile Safari (iOS) and Chrome (iOS):
- when the user clicks on the `<input>` element
  - then the `<input>` element is focused
- when the user clicks or taps outside the react-autosuggest component
  - then the `<input>` element does **NOT** lose focus

The only way to cause it to loose focus was to click on another element that could receive focus or to press the 'minimise' button on the device's keyboard.
### After this change

On Safari (Mac OS), Chrome (Mac OS, android and iOS), Mobile Safari (iOS):
- when the user clicks on the `<input>` element
  - then the `<input>` element is focused
- when the user clicks or taps outside the react-autosuggest component
  - then the `<input>` element loses focus

This replicates the behaviour of the search box on http://google.com

Also fixed bug where the user tapped or clicked on a suggestion, then dragged outside the component meaning the `mouseup` or `touchend` event wasn't triggered on the suggestion, so the `justClickedSuggestion` flag was not reset.

I tried to create some tests for this, but failed to correctly trigger events on the document.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -74,12 +74,21 @@ class Autosuggest extends Component {
     super();
 
     this.saveInput = this.saveInput.bind(this);
+    this.onDocumentTouchStart = this.onDocumentTouchStart.bind(this);
     this.onDocumentTouchEnd = this.onDocumentTouchEnd.bind(this);
+    this.onDocumentMouseUp = this.onDocumentMouseUp.bind(this);
+  }
+
+  componentWillMount() {
+    this.justTouchedInput = false;
+    this.touchStart = null;
+    this.justClickedOnSuggestion = false;
   }
 
   componentDidMount() {
-    this.touchingInput = false;
+    global.window.addEventListener('touchstart', this.onDocumentTouchStart, false);
     global.window.addEventListener('touchend', this.onDocumentTouchEnd, false);
+    global.window.addEventListener('mouseup', this.onDocumentMouseUp, false);
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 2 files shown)


src/Autosuggest.js
```diff
@@ -89,6 +104,34 @@ class Autosuggest extends Component {
...
+    this.touchStart = this.touchStart || touches[0];
+  }
+
+  onDocumentTouchEnd({ changedTouches }) {
+    const { clientX, clientY } = changedTouches[0];
+    const oldX = this.touchStart && this.touchStart.clientX || Number.MAX_VALUE;
+    const oldY = this.touchStart && this.touchStart.clientY || Number.MAX_VALUE;
+
+    if (!this.justTouchedInput && !this.justClickedOnSuggestion &&
+      Math.abs(oldX - clientX) < 20 && Math.abs(oldY - clientY) < 20) {
```
moroshko: Could you explain please what's going on here?

andysprout: My goal is to blur the input if the user taps on the page, but not blur it if the user scrolls or zooms. So I work out if the first touch object has moved more than 20px in the x or y direction since the `touchstart` event. Ideally I would bind to the `click` event, but on ios the click event is not bubbled to the document unless the target has css `cursor: pointer` http://www.quirksmode.org/blog/archives/2010/10/click_event_del_1.html

moroshko:
@andysprout Thanks a lot for this!
Any chance you could split this PR into smaller PRs, that could be reviewed and merged independently?

andysprout:
I can have a go, while trying to fix the issue that makes ios inconsistent I managed to find and fix two other bugs - I'll try to create separate prs for those bug fixes, but the ios fix will be dependant on them.

andysprout:
Created #175 I'll update this PR if that is merged in.

src/AutosuggestContainer.js
```diff
@@ -46,8 +46,8 @@ function mapToAutowhateverTheme(theme) {
   return result;
 }
 
-export default class AutosuggestContainer extends Component {
-  static propTypes = {
+export default React.createClass({
```
rickhanlonii: I'd probably keep this as a class since the React ["eventual goal is for ES6 classes to replace React.createClass completely"](https://facebook.github.io/react/blog/2015/12/18/react-components-elements-and-instances.html).

nkbt:
Rebased and reverted back to ES6 classes

ndao:
@moroshko, @nkbt, @andysprout, can somebody please help me with this? I'm having issue and this might fix it.

matheusml:
Having the same issue here @andysprout @moroshko 

When I'm on iPad, after selecting an option (```onSuggestionSelected```), the input should trigger a ```blur```, but apparently it's not. So, after selecting an option, the iPad keyboard stays there.

moroshko:
@matheusml What you are describing sounds like a different issue. Have you tried:
[focusInputOnSuggestionClick={false}](https://github.com/moroshko/react-autosuggest#focusInputOnSuggestionClickProp)
?

moroshko:
Closing for no activity.

ahmetkuslular:
@moroshko focusInputOnSuggestionClick still not working on ios devices. - 2019

Issue #158, opened 2016-05-02
Allow input blur when using keyboard for suggestion or term


JamieDixon:
Hi, first I just want to say thanks for creating and maintaining react-autosuggest. I appreciate all the effort that goes into maintaining an OSS project, 👍 

We have a prop `focusInputOnSuggestionClick` that determines whether the input receives focus when a suggestion is clicked. I am however finding it difficult to `blur` the input when the user hits `ENTER` on a suggestion or when wrapping the Autosuggest component in a `<form>` and handling the `onSubmit` event.

Is there already a recommended way of doing this or is this issue a feature request?

Thanks for any pointers and thanks again for your time and effort on this project.

moroshko:
Interesting. This is the first time someone asks to blur on Enter. Could you explain the rationale behind this? If you could create a Codepen that demonstrates the issue, that would be great.

JamieDixon:
The use case is that a user types a search query into a suggest-enabled box and hits enter (or keys down to an item and hits enter). The search is then loaded (this is an single-page-app) and because the search input still has focus, the suggestions show up.

We'd like to blur the input and stop the suggestions showing at that point because they offer little value to a customer who's already seeing search results. Perhaps this is a niche case? I could create a fork and implement this and PR if you're interested in the feature.

moroshko:
@JamieDixon Which version of react-autosuggest are you using? When you hit Enter, suggestions should be hidden like in [this Codepen](http://codepen.io/moroshko/pen/LGNJMy). Could you create a Codepen that demonstrates the issue?

JamieDixon:
Hi @moroshko Apologies in the delay in replying. I got married on Sunday so things have been quite busy!

The issue I'm having is when the autocomplete is wrapped in a form and we `preventDefault` on submit. At this point we load our search results (single-page-app) and we need the suggestions to disappear.

Here's an amended version of the pen to illustrate the issue: http://codepen.io/JamieDixon/pen/XdGrZa

Cheers!

moroshko:
@JamieDixon Congrats man!!

I just checked your pen and I see that the suggestions are hidden when you hit Enter. I thought that the issue is that they stay open. What am I missing?

JamieDixon:
Hey @moroshko Thanks for the congratulations.

In my pen, the suggestions are only hidden if you choose one of the suggested items but not when you just hit enter from the input itself. Perhaps you're seeing different behavior? I'm using Chrome 50.0.2661.94 m.

[...]

PR #171, opened 2016-05-19
Hiding autosuggest box when enter is hit and no suggestion is focused

JamieDixon:
This change hides the suggestions when the user hits enter. We use the autosuggest component within a form quite often and since suggestions are non-mandatory (since they're suggestions) we feel that the behaviour of entering a non-suggested term should be the same as picking one of the suggestions, in this case, hiding the suggestion box.

This PR relates to the discussion at: https://github.com/moroshko/react-autosuggest/issues/158

👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Check out the <a href="http://react-autosuggest.js.org" target="_blank">Homepage
 ## Installation
 
 ```shell
-npm install react-autosuggest --save
+npm install @jamiedixon/react-autosuggest --save
 ```
 
 ## Basic Usage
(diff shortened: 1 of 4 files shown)


allanesquina:
:+1:

moroshko:
@JamieDixon Thanks! I think this behaviour makes sense. I couldn't think of a use case where this behaviour wouldn't be desired.

oliverbienert:
:+1:

ccschmitz:
> we feel that the behaviour of entering a non-suggested term should be the same as picking one of the suggestions

Do you think it makes sense to actually trigger `onSuggestionSelected` in this case then? That's actually the behavior I expected.

PR #175, opened 2016-05-23
Correctly reset `justClickedOnSuggestion` on document mouseup/touchend

andysprout:
### Current behaviour

Previously `justClickedOnSuggestion` is set on `touchstart` or `mousedown` on a suggestion, but only reset when `click` events are triggered from a suggestion. If a user mousedown's or touchstart's on a suggestion, then drags outside the component, a click event is never triggered from a suggestion, so the flag remains set. To replicate the issue:
- On the demo page, click or tap the input for the the 'basic' example
- type 'c'
- mousedown or touchstart over any of the suggestions
- drag outside the suggestion list and release the mouse or your finger
- click on the 'Multiple Suggestions' input
- the 'Multiple Suggestions' input gains focus, but the suggestion list for the 'basic' example remains displayed
### Changes
- add a `onDocumentMouseUp` method that resets `justClickedOnSuggestion` on `mouseup` and `touchend` on the document.

I tried to add unit tests for this, but wasn't able to trigger events from the document using the react test utils.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -74,6 +74,12 @@ class Autosuggest extends Component {
     super();
 
     this.saveInput = this.saveInput.bind(this);
+    this.onDocumentMouseUp = this.onDocumentMouseUp.bind(this);
+  }
+
+  componentDidMount() {
+    global.window.addEventListener('mouseup', this.onDocumentMouseUp, false);
+    global.window.addEventListener('touchend', this.onDocumentMouseUp, false);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -87,6 +93,17 @@ class Autosuggest extends Component {
         revealSuggestions();
       }
     }
+  }
+
+  componentWillUnmount() {
+    global.window.removeEventListener('mouseup', this.onDocumentMouseUp, false);
+    global.window.removeEventListener('touchend', this.onDocumentMouseUp, false);
+  }
+
+  onDocumentMouseUp() {
+    setTimeout(() => {
+      this.justClickedOnSuggestion = false;
+    });
   }
 
   getSuggestion(sectionIndex, suggestionIndex) {
@@ -306,10 +323,6 @@ class Autosuggest extends Component {
       }
 
       this.maybeCallOnSuggestionsUpdateRequested({ value: clickedSuggestionValue, reason: 'click' });
-
-      setTimeout(() => {
-        this.justClickedOnSuggestion = false;
-      });
     };
     const itemProps = ({ sectionIndex, itemIndex }) => {
       return {
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -78,8 +78,8 @@ class Autosuggest extends Component {
   }
 
   componentDidMount() {
-    global.window.addEventListener('mouseup', this.onDocumentMouseUp, false);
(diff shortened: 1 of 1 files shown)


src/Autosuggest.js
```diff
@@ -74,6 +74,12 @@ class Autosuggest extends Component {
     super();
 
     this.saveInput = this.saveInput.bind(this);
+    this.onDocumentMouseUp = this.onDocumentMouseUp.bind(this);
+  }
+
+  componentDidMount() {
+    global.window.addEventListener('mouseup', this.onDocumentMouseUp, false);
+    global.window.addEventListener('touchend', this.onDocumentMouseUp, false);
```
moroshko: The name `onDocumentMouseUp` implies that it is attached to the `document`, not the `window`. Will it still do the trick if we attach it to the `document`?

src/Autosuggest.js
```diff
@@ -89,6 +95,17 @@ class Autosuggest extends Component {
...
 
+  componentWillUnmount() {
+    global.window.removeEventListener('mouseup', this.onDocumentMouseUp, false);
+    global.window.removeEventListener('touchend', this.onDocumentMouseUp, false);
+  }
+
+  onDocumentMouseUp() {
+    setTimeout(() => {
+      this.justClickedOnSuggestion = false;
+    });
```
moroshko: I'm assuming that you checked that we still need the `setTimeout` here?

moroshko:
@andysprout Thanks a lot for fixing this!

> I tried to add unit tests for this, but wasn't able to trigger events from the document using the react test utils.

[We have the jsdom `document` and `window`.](https://github.com/moroshko/react-autosuggest/blob/master/test/index.js) Have you tried to trigger the events on them?

src/Autosuggest.js
andysprout: yes, in order for `justClickedOnSuggestion` to be set when `inputProps.onBlur` is called, we still need a timeout.

src/Autosuggest.js
andysprout: Sorry this is probably a mistake where I was experimenting with using `document` or `window` to see if it had an effect and I didn't change it back.

I just did some reading on it and it seems that for mouse/touch events it makes no difference, I'll change it to bind to `global.document` to be more consistent.

andysprout:
I didn't get very far with jsdom, I was looking for something like `document.triggerEvent('mouseup')` method. I think I tried something like `document.triggerEvent(new Event('mouseup'))` but that wasn't triggering the handler. It also seemed a bit weird to trigger the 'mousedown' event using react test utils, then trigger a 'mouseup' event on jsdom's document without triggering a 'mouseup' on any dom element - it feels a bit like trying to replicate the inner workings of a browser since if the target of the two events was the same the browser would trigger a 'click' event too.

I had another thought to create another element in the test app outside the react-autosuggest component, then trigger a mouseup event on that using the test utils., But I found triggering events using the react test utils didn't get bubbled up to the document - I guess they must just stay inside the react test utils. To confirm this I added the following to the `componentDidMount` method of `Autosuggest.js`

```global.document.addEventListener('click', () => console.log('clicked'));```

Running `npm test` doesn't log any 'clicked'.

moroshko:
Closing as outdated.

PR #176, opened 2016-05-24
Use custom host and port

andysprout:
In order to test on other devices on my LAN I needed to be able to set the host to my ip. eg if my ip was `10.4.1.19`, then run:

```NODE_HOST=10.4.1.19 npm start```

Then navigate to http://10.4.1.19:3000/demo/dist on another device (eg ipad or smart phone) on your LAN

`npm start` continues to work on http://localhost:3000


moroshko:
Thanks @andysprout!

PR #188, opened 2016-06-10
Simple improvement at workflow of development

cauealves:
This is a simple change to open the URL after creating the server with `npm start` command.


rickhanlonii:
@moroshko @cauealves what do you think about putting this behind an `--open` flag of some sort rather than default behavior?

cauealves:
@rickhanlonii It's possible. I'm not sure if this is really necessary. This is a common flow to those who are raising the server.. But it's a case to think about =) 

What did you think about this, @moroshko ?

moroshko:
@cauealves I agree that most of the time you'd want to open the page manually, so why not to automate this :) Mind resolving the conflicts?

cauealves:
@moroshko Of course. Now, everything is up to date =)

moroshko:
Thanks!

PR #192, opened 2016-07-05
Update code to use bindActionCreators provided by redux

mauriciosoares:
Hey there,

In this PR i simply refactor the `mapDispatchToProps` method to use `bindActionCreators` provided by the version of redux you're already using. It cleans up the code a little bit.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -1,7 +1,7 @@
 import React, { Component, PropTypes } from 'react';
+import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


moroshko:
Thanks! It indeed makes the code a bit cleaner 👏

moroshko:
@mauriciosoares I just discovered that you can pass `actionCreators` instead of `mapDispatchToProps` to `connect()`, so I [updated](https://github.com/moroshko/react-autosuggest/commit/9481ea7eeb7075880eb5b74004d238ee9d0d5dd3) the code accordingly.

mauriciosoares:
wow! that`s pretty cool! I didn`t know that.

Thanks for telling me :metal:

PR #235, opened 2016-09-01
Example function name mismatch

tomsowerby:
Looks like the quickstart should be using the onSuggestionsFetchRequested function?


moroshko:
Thanks for pointing out!

tomsowerby:
You're welcome, thanks for the library.

PR #261, opened 2016-10-05
switch onChange to onInput because of IE11 react bug

cuth:
The reason for this PR is because of this React Bug: https://github.com/facebook/react/issues/7027

Instead of waiting for React v16, we are editing this component to use onInput instead of onChange to see if that helps IE11.

This breaks all the tests and just replacing "onChange" with "onInput" didn't fix the tests.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -158,10 +158,10 @@ class Autosuggest extends Component {
   }
 
   maybeCallOnChange(event, newValue, method) {
-    const { value, onChange } = this.props.inputProps;
+    const { value, onInput } = this.props.inputProps;
 
     if (newValue !== value) {
-      onChange && onChange(event, { newValue, method });
+      onInput && onInput(event, { newValue, method });
     }
   }
 
@@ -308,7 +308,7 @@ class Autosuggest extends Component {
           this.onSuggestionsClearRequested();
         }
       },
-      onChange: event => {
+      onInput: event => {
         const { value } = event.target;
         const shouldRender = shouldRenderSuggestions(value);
 
--- a/src/AutosuggestContainer.js
+++ b/src/AutosuggestContainer.js
@@ -74,8 +74,8 @@ export default class AutosuggestContainer extends Component {
         throw new Error('\'inputProps\' must have \'value\'.');
       }
 
-      if (!inputProps.hasOwnProperty('onChange')) {
-        throw new Error('\'inputProps\' must have \'onChange\'.');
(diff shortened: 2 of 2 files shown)


cuth:
So sorry, my bad, I only wanted to do this locally!

jbandi:
I would be interested if this resulted in an acceptable workaround until React 16 is released?

cuth:
So far, so good

tannerjt:
I was able to implement the change (onChange => onInput), however the problem persists.  Is this workaround actually binding the method to the onInput event, or just changing the name of the property that is passed on?  I am thinking the workaround would actually need to be done on react-autowhatever, since that is the component that creates the input element.  My fingers are crosses for react 16 to come out soon.

jbandi:
We are using this implementation of react-autosuggest in our project:
https://github.com/postlight/react-autosuggest

The bug described here: https://github.com/moroshko/react-autosuggest/issues/262
is fixed in that fork of react-autosuggest.

tannerjt:
Thanks @jbandi.  It would appear it isn't merged into the master branch, correct?  I'm assuming the correct branch with the edits is the fix-onchange-to-oninput?  I thought I had tried this and was still getting errors in IE11.  I'll clone your fork and try again.

tannerjt:
The fix isn't working for me.  Are you testing in IE11?  I think your changes ([here](https://github.com/postlight/react-autosuggest/commit/77ee4e997312929f3b31f5f1d7247b54be95795c)) are only biding a new function called onInput that is secretly still biding to the native onChange event.  I'll probably wait until react 16 comes out, since this seems to be a problem with React

jbandi:
I am not sure if we are talking about the same problem.

In our project we had the following problem:
https://github.com/moroshko/react-autosuggest/issues/262

The above problem is fixed with the following fork of react-autosuggest:
https://github.com/postlight/react-autosuggest

We are testing in IE11.

I did not really look at the implementation of the fix. But as seen in the bug report here:
https://github.com/facebook/react/issues/7027

... the underlying problem seems to come from "wrong" timing or the OnChange Event, which is a synthetic event of React. Therefore switching to the OnInput Event could well be the fix...

JacksonKearl:
For a version with @cuth 's fix, which also passes the tests: see my [branch](https://github.com/JacksonKearl/react-autosuggest-ie11-compatible).

Or `npm install react-autosuggest-ie11-compatible`. This is a direct drop in replacement for moroshko's version; there's no need to mess around with changing `onInput` or `onChange` in the invocation.

PR #263, opened 2016-10-10
Added inputElement support from react-autowhatever 5.4.0

angelo-v:
This pull request adds the support for a custom input element, that came with react-autowhatever 5.4.0 to react-autosuggest.


angelo-v:
The bitHound lint result (indentation should be 8) contradicts the result of "npm run lint" (indentation should be 6). Which one is correct?

src/Autosuggest.js
```diff
@@ -440,6 +444,7 @@ class Autosuggest extends Component {
         inputProps={autowhateverInputProps}
         itemProps={this.itemProps}
         theme={theme}
+        inputElement={this.props.inputElement}
```
moroshko: If I understand correctly, we'll be passing `undefined` here by default, relying on `react-autowhatever` to have an `'input'` fallback.

What do you think about being more explicit by:
* making `inputElement ` required in `Autosuggest.js`
* default `inputElement` to `'input'` in `AutosuggestContainer.js`

?

moroshko:
@angelo-v The lint issue is probably [a bug in `eslint-plugin-react`](https://github.com/yannickcr/eslint-plugin-react/issues/901). I guess we can lock its version to `6.2.2` until it's fixed.

Could you also please update the docs to reflect the changes, and add tests?

src/Autosuggest.js
angelo-v: Sounds fine for me, I adjusted it accordingly

moroshko:
Thanks!

moroshko:
@angelo-v I ended up renaming `inputElement` to `inputComponent` and constraining it to a React Component (cannot be string like `'textarea'`) to keep Autosuggest accessible.

It is available in [v6.1.0](https://github.com/moroshko/react-autosuggest/releases/tag/v6.1.0) now.

angelo-v:
Ok, great! Thanks for merging!

PR #265, opened 2016-10-10
Updated UPGRADE_GUIDE for 5.x to 6.x

zachhale:
While upgrading from 5.x to 6.x I scoured the release log and found some key information that should exist in the upgrade guide. I copied it over.


moroshko:
Thanks @zachhale

PR #270, opened 2016-10-19
Fix Uncaught TypeError: Cannot read property 'trim' of undefined

ChrisCinelli:
In one input I use react-autosuggest to display suggestions BEFORE the user start typing.
If this case if the user press `down`, I get:
```
Uncaught TypeError: Cannot read property 'trim' of undefined
```

This small change  the problem.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -427,7 +427,7 @@ class Autosuggest extends Component {
       }
     };
     const renderSuggestionData = {
-      query: (valueBeforeUpDown || value).trim()
+      query: (valueBeforeUpDown || value || "").trim()
     };
 
     return (


moroshko:
@ChrisCinelli I tried to reproduce the error you are getting [here](http://codepen.io/moroshko/pen/kXpapq), but no luck.

Which version of `react-autosuggest` are you using?

Could you please create a Codepen that demonstrates the issue?

moroshko:
Closing for now, but happy to reopen once a reproducible Codepen is provided.

hughlomas:
I'm having the same issue on `react-autosuggest@7.0.1`, with the same line of code as ChrisCinelli, when first typing into the input. 

`Uncaught TypeError: Cannot read property 'trim' of undefined`

```javascript
var renderSuggestionData = {
  query: (valueBeforeUpDown || value).trim()
};```

Sorry but I don't have a codepen to reproduce and don't have time to make one right now. If I discover the cause I will update this.

mr-devboy:
version 9.4.2, same error

Issue #281, opened 2016-11-07
Using with <textarea>


Gargron:
So, *mostly*, this component works with a textarea. However, there are some behaviours that would be nice to be changed when it's used like that:

- Up/down arrow keys should not be a no-op if suggestions are empty
- Tab key should be the same as Enter

Can we add options for this?

moroshko:
Could you create Codepen with your mostly working example?
Also, I'm not sure I understand the first point. Could you provide an example and describe the expected behaviour please?

Gargron:
For whatever reason, all Codepen examples only work in Chrome, not in Firefox (though the component itself works fine in Firefox, I think it's just a Codepen issue). Here is an example: <https://codepen.io/anon/pen/qqORdp>

Note how up/down keys are reserved even when suggestions are empty, also how "enter" to select suggestion doesn't work, you have to click it.

Gargron:
I haven't tested it but I think perhaps "enter" doesn't work because normally an `onChange` handler is also attached to the input, and they fire at the same time and the `setState` from that overrides the change from `onSuggestionSelected`, so perhaps what we need to do is stop propagation when enter (or tab) is pressed *and* suggestions are not empty.

Similarly, simply don't prevent default behaviour on up/down arrows when suggestions are empty. I don't think that would break <input>s either.

a-x-:
will be cool decouple textinput and make possible to use any external texteditor

a-x-:
for example i want use this awesome suggest component with the facebook «Draft» texteditor

moroshko:
@a-x- I like the idea of decoupling, but not sure what would the (new decoupled) API look like. If you have any ideas, please open a new issue with your proposal!

moroshko:
Closing for now.

oyeanuj:
@a-x- Did you end up using it with Draft?

@moroshko I was looking to use it with [Slate editor](https://github.com/ianstormtaylor/slate/) for mentions and other auto-complete behavior. I think the key complication in the current design is that the `input` and `autosuggest` belong to the same container parent, and there are props being passed to the `input` In the case of RTEs, that is hard since you need some child node components to have the autosuggest. Additionally, `onChange` and other such handlers are being used in some other manner.

[...]

PR #289, opened 2016-11-11
Feature textarea

Gargron:
See #281. This PR changes the behaviour slightly, which is why I had to disable one test case for old behaviour, hopefully you can figure out if it's something you'd want to adjust or remove.

Changes:

- Do not suppress up/down arrow key presses when no suggestions are loaded (in-textarea navigation).
- Add the same behaviour to <kbd>Tab</kbd> key as <kbd>Enter</kbd> key.
- Suppress <kbd>Tab</kbd>/<kbd>Enter</kbd> key presses when a suggestion is being selected (aka prevent line break/focus switch when you're selecting a suggestion).
- Remove call to `maybeCallOnChange` from <kbd>Tab</kbd>/<kbd>Enter</kbd> key press, because otherwise whatever the `onSuggestionSelected` handler does will be overwritten by the value passed to `maybeCallOnChange` immediately.

Furthermore, a textarea example is added to the demo.

### DIFF
--- a/demo/src/components/App/components/Examples/Examples.js
+++ b/demo/src/components/App/components/Examples/Examples.js
@@ -5,6 +5,7 @@ import Basic from 'Basic/Basic';
 import MultipleSections from 'MultipleSections/MultipleSections';
 import CustomRender from 'CustomRender/CustomRender';
 import ScrollableContainer from 'ScrollableContainer/ScrollableContainer';
+import TextareaRender from 'TextareaRender/TextareaRender';
 
 const Examples = () => (
   <div className={styles.container}>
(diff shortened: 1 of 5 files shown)


Gargron:
Could this get reviewed please?

moroshko:
@Gargron Sorry, I'm very busy lately, but I'll try to find time soon to review this PR.

adrianmcli:
@Gargron looks like my recent PR #305 caused a conflict in your PR. I'm curious why you decided to comment out the `maybeCallOnChanged()` line?

Gargron:
I couldn't wait for the PR any longer and re-did the entire autosuggestions component for textareas in my own code. For the curious: https://github.com/Gargron/mastodon/blob/master/app/assets/javascripts/components/components/autosuggest_textarea.jsx

This PR might as well be closed.

PR #298, opened 2016-12-05
README: Fixed id link to onSuggestionSelected

zacacollier:
Corrected id link for onSuggestionSelected. All other id links I tested are working properly.


moroshko:
Are you sure the link isn't working? It works for me.

zacacollier:
Double-checked and indeed the original link works as it should - not sure why it didn't work for me earlier. Browsing too fast perhaps? Anyway, thanks for giving it a look!

PR #306, opened 2016-12-15
Call onSuggestionSelected after onChange

### LINKED ISSUE #305: When hitting enter to select a suggestion, the input does not clear.
This is a very specific bug. It occurs when:

  1. `focusFirstSuggestion` is set to `true`.
  2. The input is to be cleared when a suggestion is selected.
    - i.e. `onSuggestionSelected()` contains `this.setState({ value: '' });`.
  3. You hit <kbd>Enter</kbd> to select the first suggestion that was automatically focused onto.
  4. The first suggestion must not match exactly the thing that was typed in.

### Demo

This pen has two major differences with the basic one:

  1. `focusFirstSuggestion` is set to `true`.
  2. A function is passed in into the `onSuggestionSelected` prop:

  ```js
    onSuggestionSelected = () => {
      this.setState({ value: '' })
    };
  ```

Try it [here](http://codepen.io/anon/pen/dOQRgZ?editors=0010).

  1. Focus on the input field.
  2. Type `c` (lowercase), and wait for suggestions to appear.
  3. Press <kbd>Enter</kbd> to select the first suggestion.

  **Observed behaviour:**

Note that the `c` remains, and the input is not cleared.

  **Expected behaviour:**

The input should be cleared, since there is already a focus on the first item of the suggestions. And we have `onSuggestionSelected` set to clear the input for us. Unfortunately, `onChange` is being called after `onSuggestionSelected` when it should not be.

### The Problem

When the suggestion is selected by hitting <kbd>Enter</kbd>, the `onSuggestionSelected` function is called, and the input is properly cleared. **However**, immediately after that, `maybeCallOnChange` is called.

[Line 369 to 376](https://github.com/moroshko/react-autosuggest/blob/6f3d2257515a3cc71c5a96b0eab2ebf866444339/src/Autosuggest.js#L369-L376):

```js
this.onSuggestionSelected(event, {
  suggestion: focusedSuggestion,
  suggestionValue: newValue,
  sectionIndex: focusedSectionIndex,
  method: 'enter'
});

this.maybeCallOnChange(event, newValue, 'enter');
```

This takes us to the definition on [line 160](https://github.com/moroshko/react-autosuggest/blob/6f3d2257515a3cc71c5a96b0eab2ebf866444339/src/Autosuggest.js#L160-L166) for `maybeCallOnChange()`:

```js
maybeCallOnChange(event, newValue, method) {
  const { value, onChange } = this.props.inputProps;

  if (newValue !== value) {
    onChange(event, { newValue, method });
  }
}
```

Since the current value was typed in by the user, and the `newValue` came from the selected suggestion, `onChange` is called with the `newValue`. This leads us to our simple `onChange` function:

```js
onChange = (event, { newValue, method }) => {
  this.setState({ value: newValue });
}
```

As you can see, it just sets the value to the state and you end up with the new suggestion in the input rather than a cleared box.

### The (temporary) Solution

In the `onChange` handler function that we pass down to the Autosuggest component, we must first check if the method is `enter`. If it is, then prevent the function from continuing:

```js
onChange = (event, { newValue, method }) => {
  // prevent change if artist was just selected
  if (method === 'enter') return;
  // otherwise, let the change go through
  this.setState({ value: newValue });
}
```

### A Permanent Solution

We need to prevent `onChange` from being called after `onSuggestionSelected` when:

  1. `focusFirstSuggestion` is set to `true`, and;
  2. The user pressed <kbd>Enter</kbd> to trigger `onSuggestionSelected`.

adrianmcli:
This PR fixes #305, please read the issue for a more in-depth look at the problem.

In that issue, I suggested that we should prevent the call to `onChange` if the submission method was `Enter` KeyDown. After some thinking and reading of the code, it makes a lot more sense to simply move the `maybeCallOnChange` call above the `onSuggestionSelected` call instead.

The idea is that we want `onSuggestionSelected` to take precedence over `onChange`, since the user's ultimate intention is to select the suggestion.

This is exactly what is done for the `click` action already. So it makes sense that the `enter` action should behave similarly. This PR simply brings it in line with that behaviour.

Note: [Lines 228 to 234](https://github.com/moroshko/react-autosuggest/blob/master/src/Autosuggest.js#L228-L234) is where this happens for the click event.

```js
this.maybeCallOnChange(event, clickedSuggestionValue, 'click');
this.onSuggestionSelected(event, {
  suggestion: clickedSuggestion,
  suggestionValue: clickedSuggestionValue,
  sectionIndex,
  method: 'click'
});
```

I've made two tests for the feature. Note that I had to create a new folder for the test because I needed a `onSuggestionSelected` function that clears the input value with `this.setState({ value: '' })`.

This PR passes all tests, linting, and the build process.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -366,14 +366,13 @@ class Autosuggest extends Component {
             if (focusedSuggestion !== null) {
               const newValue = getSuggestionValue(focusedSuggestion);
 
+              this.maybeCallOnChange(event, newValue, 'enter');
               this.onSuggestionSelected(event, {
                 suggestion: focusedSuggestion,
                 suggestionValue: newValue,
                 sectionIndex: focusedSectionIndex,
                 method: 'enter'
               });
-
-              this.maybeCallOnChange(event, newValue, 'enter');
 
               this.justSelectedSuggestion = true;
 
--- a/test/focus-first-suggestion-clear-on-enter/AutosuggestApp.js
+++ b/test/focus-first-suggestion-clear-on-enter/AutosuggestApp.js
@@ -1,0 +1,80 @@
+import React, { Component } from 'react';
+import sinon from 'sinon';
+import Autosuggest from '../../src/AutosuggestContainer';
+import languages from '../plain-list/languages';
+import { escapeRegexCharacters } from '../../demo/src/components/utils/utils.js';
+
+const getMatchingLanguages = value => {
+  const escapedValue = escapeRegexCharacters(value.trim());
+  const regex = new RegExp('^' + escapedValue, 'i');
+
+  return languages.filter(language => regex.test(language.name));
+};
+
+let app = null;
+
+export const getSuggestionValue = sinon.spy(suggestion => {
+  return suggestion.name;
+});
+
+export const renderSuggestion = sinon.spy(suggestion => {
+  return (
+    <span>{suggestion.name}</span>
+  );
+});
+
+export const onChange = sinon.spy((event, { newValue }) => {
+  app.setState({
+    value: newValue
+  });
+});
+
+export const onSuggestionsFetchRequested = sinon.spy(({ value }) => {
+  app.setState({
+    suggestions: getMatchingLanguages(value)
+  });
+});
+
+export const onSuggestionsClearRequested = sinon.spy(() => {
+  app.setState({
+    suggestions: []
+  });
+});
+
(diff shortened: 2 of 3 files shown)


adrianmcli:
@moroshko not sure why the CI checks have failed? Everything looks good if you go to the details.

This is a really tiny PR, all I did was move a piece of code a few lines up.

PR #311, opened 2016-12-31
Pass suggestionIndex to onSuggestionSelected

oyeanuj:
Hi @moroshko! 

I’ve needed the `suggestionIndex` of the selected suggestion a couple of times in my application. 

So, I figured I would make a quick PR to see if this is something you’d like to add to the library.

Thank you for providing this library!

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,7 +344,7 @@ where:
 This function is called when suggestion is selected. It has the following signature:
 
 ```js
-function onSuggestionSelected(event, { suggestion, suggestionValue, suggestionIndex,sectionIndex, method })
+function onSuggestionSelected(event, { suggestion, suggestionValue, suggestionIndex, sectionIndex, method })
 ```
 
 where:
--- a/README.md
+++ b/README.md
@@ -351,6 +351,7 @@ where:
 
 * `suggestion` - the selected suggestion
 * `suggestionValue` - the value of the selected suggestion (equivalent to `getSuggestionValue(suggestion)`)
+* `suggestionIndex` - the index of the selected suggestion in the `suggestions` array )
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -351,6 +351,7 @@ where:
 
 * `suggestion` - the selected suggestion
 * `suggestionValue` - the value of the selected suggestion (equivalent to `getSuggestionValue(suggestion)`)
+* `suggestionIndex` - the index of the selected suggestion in the `suggestions` array )
```
moroshko: The `)` at the end should probably not be there. Also, `suggestionIndex` need to be added to the signature few lines above.

oyeanuj: Fixed in the latest commit.

oyeanuj:
So, I had two different use-cases - 

1. I needed to get the original search result from the redux store, and having the key was a little more convenient than finding by value.

2. To get some usability insights into how people are finding and using the results, it is useful to log how relevant are the search results to the term the users entered.

I figured other people might find other uses as well, so worth putting it out there.

Let me know what you think!

README.md
```diff
@@ -344,13 +344,14 @@ where:
 This function is called when suggestion is selected. It has the following signature:
 
 ```js
-function onSuggestionSelected(event, { suggestion, suggestionValue, sectionIndex, method })
+function onSuggestionSelected(event, { suggestion, suggestionValue, suggestionIndex,sectionIndex, method })
```
moroshko: Please add a space before `sectionIndex` to keep a consistent style.

oyeanuj: My bad, should be fixed now.

moroshko:
It's in [v7.1.0](https://github.com/moroshko/react-autosuggest/releases/tag/v7.1.0) now.

PR #322, opened 2017-02-07
Enforce node != document checks

lPadier:
Check that node is not document nor null on first iteration.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -265,7 +265,7 @@ export default class Autosuggest extends Component {
       event.detail && event.detail.target || // This is for testing only. Please show me a better way to emulate this.
       event.target;
 
-    do {
+    while (node !== null && node !== document) {
       if (node.getAttribute('data-suggestion-index') !== null) {
         // Suggestion was clicked
         return;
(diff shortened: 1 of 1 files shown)


moroshko:
Could you explain the problem you are trying to solve?

lPadier:
In the app I'm building, `node.getAttribute('data-suggestion-index')` sometimes fails on IE 11 with the following error message: "object does not support property or method getAttribute". Checking that node !== document or null before the first loop iteration should avoid event.target being document

moroshko:
Thanks!
Did you test that this change solves the issue in IE11?

lPadier:
Dispatching a mousedown event on document with this change no longer throws an error on IE11

moroshko:
[v8.0.1](https://github.com/moroshko/react-autosuggest/releases/tag/v8.0.1) contains this fix. Thanks!

Degee:
Actually i am getting exactly the same error with version 9.3.2 of autosuggest 

`object does not support property or method getAttribute`

I think this was not really a solution. Weird thing is that error sometimes randomly dissappears.

PR #331, opened 2017-03-01
Fix error when setting focusInputOnSuggestionClick to false

### LINKED ISSUE #275: Problem with focusInputOnSuggestionClick and Redux Form
I'm using `focusInputOnSuggestionClick={ false }`.
But when I select an item I see the following error:
```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the method `stopPropagation` on a released/nullified synthetic event. This is a no-op function. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.
```
one time for each of these four properties: `stopPropagation` (above), `preventDefault`, `nativeEvent` and `target`.

And then the follwing error:
```
Uncaught TypeError: Cannot read property 'type' of null
```

That last error occurs in `redux-form.js` inside the `getValue` function.

```javascript
var getValue = function getValue(event, isReactNative) {
	if ((0, _isEvent2.default)(event)) {
		if (!isReactNative && event.nativeEvent && event.nativeEvent.text !== undefined) {
			return event.nativeEvent.text;
		}
		if (isReactNative && event.nativeEvent !== undefined) {
			return event.nativeEvent.text;
		}
		var _event$target = event.target;
		var type = _event$target.type;
		var value = _event$target.value;
		var checked = _event$target.checked;
		var files = _event$target.files;
		var dataTransfer = event.dataTransfer;

		if (type === 'checkbox') {
			return checked;
		}
		if (type === 'file') {
			return files || dataTransfer && dataTransfer.files;
		}
		if (type === 'select-multiple') {
			return getSelectedValues(event.target.options);
		}
		if (value !== '' && (type === 'number' || type === 'range')) {
			return parseFloat(value);
		}
		return value;
	}
	return event;
};
```

I'm not exactly sure why, but if I comment out the line ` _this2.blurEvent = event;` in the `onBlur` method in `Autosuggest.js` the problem goes away and everything works.

I've also tried first calling `event.persist()` on the event object. The error goes away too but then the input does not get updated with the new value.

If you need me to test some things out or log some values etc. please let me know.

chris-verclytte:
Fix error when setting focusInputOnSuggestionClick to false and using component 
with HOC (like Redux Form) listening for onBlur callback.

Fixes #275.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-autosuggest",
-  "version": "9.0.0",
+  "version": "9.0.1",
   "description": "WAI-ARIA compliant React autosuggest component",
   "main": "dist/index.js",
   "repository": {
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -422,6 +422,7 @@ export default class Autosuggest extends Component {
           return;
         }
 
+        event.persist();
         this.blurEvent = event;
 
         if (!this.justSelectedSuggestion) {


moroshko:
Is the issue you are trying to solve reproducible in a browser, or is it a mobile only issue?
Could you please provide the steps to reproduce the issue, and create a failing test?

moroshko:
No activity, closing for now.

Azein:
Encountered this problem, I'm using Redux Form and custom input styled component. Seems like this PR could fix the issue.
Redux Form guidelines recommend to use e.preventDefault() to not trigger onBlur when needed - and I did exactly this, got the error. For now I have to approach this without using event and it's sad.
Reopen/merge please, have a good day.

Azein:
Anyway, working solution is to just pass some blur event handler down which returns nothing (and thus not triggering Redux Form BLUR action).  If event.persist() solution is being considered as overhead, this approach should probably be mentioned in docs/guidelines.

PR #343, opened 2017-03-17
Fix issue Edge win 10 check node getAttribute

karlAlnebratt:
Clicking on an SVG element in Edge IE 10 produces:

> Object doesn't support property or method 'getAttribute'

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -230,7 +230,7 @@ export default class Autosuggest extends Component {
       event.target;
 
     while (node !== null && node !== document) {
-      if (node.getAttribute('data-suggestion-index') !== null) {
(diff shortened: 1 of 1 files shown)


moroshko:
Thanks for submitting this PR.
If we just do what you suggested, the functionality will not be the same when `getAttribute` is not defined.
Is this a known issue in Edge? Could you provide some pointers?
I wonder if we need to find a proper workaround to keep the same functionality.

moroshko:
No activity, closing for now.

mvirbicianskas:
Any progress on this? I encountered the same problem with IE 14

BorodinDK:
I get this problem with IE 11

Issue #351, opened 2017-04-12
Support React@15.5


taylor-smith:
A number of deprecation warnings were added in React 15.5.0. Can react-autosuggest support the new recommended APIs?

https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

moroshko:
Submit a PR?

taylor-smith:
Will do! Thanks

moroshko:
Solved in [v9.0.1](https://github.com/moroshko/react-autosuggest/releases/tag/v9.0.1)

PR #352, opened 2017-04-14
React 15.5.0 compatibility

taylor-smith:
This should prevent deprecation warnings from being thrown when React 15.5.0 is used. A warning is displayed when running `npm run test` but I suspect that this is due to a dependency (mocha, perhaps?) rather than due to an introduction in this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "prepublish": "npm run dist && npm run standalone"
   },
   "dependencies": {
+    "prop-types": "^15.5.8",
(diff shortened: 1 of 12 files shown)


brendoncrawford:
This works for me.  Thanks for doing this.   The `react-autowhatever` package still needs to be updated as well.  I will do that in my own fork.

gido:
Related to issue #351

moroshko:
Thanks!

PR #385, opened 2017-06-13
Add `reason` to `onSuggestionsFetchRequested`

### LINKED ISSUE #382: Query - How to get full list on Focus
Consider the following use-case:

1)  The AutoSuggest has a value in the input
2) focus is lost and the user does something else
3) The user clicks on the AutoSuggest
4) The AutoSuggest opens
5) `_onSuggestionsFetchRequested` is called which, in my case, syncronously updates the suggestions based on the input.

`_onSuggestionsFetchRequested` is also called when you type into the input, as expected.

However, there is a subtle difference in how you want the suggestions to be rendered if you give focus to the AutoSuggest vs. typing into it.

If I give focus into an autosuggest, if it has a value in there, it would be nice to show, maybe all, or a bigger set of suggestions.  The user wants to choose a new value, so it would be good to show them a bigger list.  However, if a value is there already that scopes the suggestions down to the single value, they then have to delete text.

At the moment it looks like there is no hook for me to determine if it was the user giving the control focus or just typing into the auto suggest so that I can give more hints to the suggestion builder.

Any thoughts?

raybooysen:
Supplies a reason to the onSuggestionsFetchRequested to give the consumer more understanding on why this has been requested.

Fixes #382

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,10 +240,18 @@ const suggestions = [
 This function will be called every time you need to update [`suggestions`](#suggestionsProp). It has the following signature:
 
 ```js
-function onSuggestionsFetchRequested({ value })
+function onSuggestionsFetchRequested({ value, reason })
 ```
 
-where `value` is current value of the input.
+where
+
+ - `value` is current value of the input.
+ -  `reason` is one of the following values: 
+   - 'escape-pressed' 
+   - 'input-changed' 
+   - 'input-focused'
+   - 'suggestions-revealed'
+   - 'suggestion-selected'
 
 <a name="onSuggestionsClearRequestedProp"></a>
 #### onSuggestionsClearRequested (required unless `alwaysRenderSuggestions={true}`)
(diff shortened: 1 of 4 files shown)


demo/src/components/App/App.js
```diff
@@ -6,13 +6,12 @@ import Features from 'Features/Features';
 import Examples from 'Examples/Examples';
 import Footer from 'Footer/Footer';
 
-const App = () => (
+const App = () =>
```
raybooysen: These changes all come from running prettier

moroshko:
@raybooysen Sorry to bother you again. Could you please rebase on master, which has the latest version of prettier? All these styling changes should disappear.

raybooysen:
Done.  Way better now. :)

moroshko:
`reason` in available now is [v9.2.0](https://github.com/moroshko/react-autosuggest/releases/tag/v9.2.0).

Issue #368, opened 2017-05-21
Uncaught TypeError: Cannot read property 'trim' of undefined


jopcode:
When i select a item 

When i type in the search bar this show me the results, but when i click on a value the console give me this 'Uncaught TypeError: Cannot read property 'trim' of undefined', and my app don't reload the view until I deleted the value on search bar. Any idea or solution? ( sorry for my bad english)

Screenshots:

After type on the search bar:

![image](https://cloud.githubusercontent.com/assets/16161754/26285672/a397828a-3e54-11e7-8eb9-710d24ea062d.png)

When i type on search bar style (doesn't matter)
![image](https://cloud.githubusercontent.com/assets/16161754/26285688/d5639114-3e54-11e7-9205-30b23ad3ff44.png)

And when i click on a value:
![image](https://cloud.githubusercontent.com/assets/16161754/26285700/f9919d60-3e54-11e7-870b-06d3ae8deb9e.png)

When i deleted the value on search bar:
![image](https://cloud.githubusercontent.com/assets/16161754/26285711/2370ec80-3e55-11e7-96fe-fa5b4ec41f5b.png)

First TypeError:
![image](https://cloud.githubusercontent.com/assets/16161754/26285713/3c56e7d6-3e55-11e7-9651-bb82703462d2.png)

And the last TypeError:
```
shouldRenderSuggestions	@	AutosuggestContainer.js:174
componentWillReceiveProps	@	Autosuggest.js:72
(anonymous function)	@	ReactCompositeComponent.js:611
measureLifeCyclePerf	@	ReactCompositeComponent.js:75
updateComponent	@	ReactCompositeComponent.js:610
receiveComponent	@	ReactCompositeComponent.js:547
receiveComponent	@	ReactReconciler.js:125
_updateRenderedComponent	@	ReactCompositeComponent.js:754
_performComponentUpdate	@	ReactCompositeComponent.js:724
updateComponent	@	ReactCompositeComponent.js:645
receiveComponent	@	ReactCompositeComponent.js:547
receiveComponent	@	ReactReconciler.js:125
_updateRenderedComponent	@	ReactCompositeComponent.js:754
_performComponentUpdate	@	ReactCompositeComponent.js:724
updateComponent	@	ReactCompositeComponent.js:645
receiveComponent	@	ReactCompositeComponent.js:547
receiveComponent	@	ReactReconciler.js:125
updateChildren	@	ReactChildReconciler.js:109
[...]

PR #391, opened 2017-06-24
Convert non-string values to empty strings so that trim() doesn’t throw a TypeError

timothyjellison:
When null values are passed into the Autosuggest component's state I receive a TypeError originating from one of the two functions where the String.prototype.trim() method is used (defaultShouldRenderSuggestions and getQuery). My PR provides both functions with type checking and a fallback empty string to make sure the value being trimmed is a string.

Relevant issues: 
https://github.com/moroshko/react-autosuggest/issues/368

* npm run build was completed successfully

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -5,7 +5,11 @@ import Autowhatever from 'react-autowhatever';
 import { defaultTheme, mapToAutowhateverTheme } from './theme';
 
 const alwaysTrue = () => true;
-const defaultShouldRenderSuggestions = value => value.trim().length > 0;
+const defaultShouldRenderSuggestions = value => {
+  const formattedValue = typeof value === 'string' ? value.trim() : '';
+
+  return formattedValue.length > 0;
+};
 const defaultRenderSuggestionsContainer = ({ containerProps, children }) =>
   <div {...containerProps}>{children}</div>;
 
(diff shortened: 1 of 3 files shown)


timothyjellison:
@moroshko if you have any comments or suggestions on changes I'd be happy to hear them. This is my first PR on an open source project so apologies if I missed protocol somewhere.

moroshko:
Hey Tim,

Sorry, it's not you, it's me.
I've been swamped recently and barely find time to look at Pull Requests.

The main concern I have with this fix is that it silences a real issue.
For the Autosuggest to work correctly, `value` needs to be a string.
If it's not, the user needs to know about this as soon as possible.
I prefer to fail with an error immediately.

The problem though, is that currently the error message is not descriptive, and doesn't guide the user towards the fix. The following message would be of more value to the user:
> To work correctly, Autosuggest requires input's value to be a string. Found: null

Would you like to fix the Pull Request to throw an error instead?

timothyjellison:
Hey no worries, and thanks for getting back to me. 😄  

I see your point; I'll update the PR to throw an error rather than just eat the null silently. Should have time to do it over the next couple days.

moroshko:
I just updated `master` with the latest version of Prettier.
Could you please rebase on `master` so that no Prettier changes are included in this PR?

Also:
* Did you forget to throw the error in `getQuery()`? (we `trim()` there as well)
* Could you please add a test instead of lowering the coverage threshold?

Issue #406, opened 2017-07-20
Old broswers and keyboard problem


AlgoTrader:
You have the following code in Autosuggest:
```
          switch (event.key) {
            case 'ArrowDown':
            case 'ArrowUp':
```

The problem is old browsers (Сhrome 48 in the case) have no `event.key`: 
![image](https://user-images.githubusercontent.com/1556262/28415492-b729250a-6d58-11e7-855d-c7aa5834da25.png)

Can you please rewrite the code like this:
```
          const keyCode = event.key || event.code;
          switch (keyCode) {
            case 'ArrowDown':
            case 'ArrowUp':
```
It will fix keyboard operation in Chrome 48 that is just 2 years old

moroshko:
Could you submit a PR please?

AlgoTrader:
Done. Please boost the npm version, my boss has Chrome 48 :(

moroshko:
[v9.3.2](https://github.com/moroshko/react-autosuggest/releases/tag/v9.3.2) is on npm.

I ended up using `event.keyCode` instead of `event.key || event.code`.

Could you please verify that v9.3.2 solves your issue on Chrome 48?

PR #407, opened 2017-07-20
Fix the keydown handling in old browsers

AlgoTrader:
Fixes the keyboard operation in many old browsers, tested in Chrome 48.
See https://github.com/moroshko/react-autosuggest/issues/406 for details


AlgoTrader:
It is a plain fallback from `event.key` to `event.code` when `event.key` is missing in keyboard event

Issue #380, opened 2017-06-05
inputProps.onBlur not working on mobile


jpap:
* Try out the [homepage](http://react-autosuggest.js.org/) or [codepen](http://codepen.io/collection/DkkYaQ) examples on mobile (iOS or Android).  No issue on desktop browsers.
* Steps to reproduce:
  1. Select the component, type a character to open the dropdown suggestion list.
  1. Select any item in the dropdown; e.g. scroll the list if possible.
  1. Tap elsewhere on the page (e.g. another dropdown).
  1. Observe that the original component doesn't "close" (dropdown still open).
* **Expected behavior:** The dropdown should have closed (and `inputProps.onBlur` event fired) when another component is selected.

moroshko:
This is a known issue. 
Previous attempts to fix this [here](https://github.com/moroshko/react-autosuggest/pull/170).
PR is welcome!

scotta-koorong:
Hey all,

The issue lies in justSelectedSuggestion blocking the .blur() on the element. If justSelectedSuggestion is not set to false, then it doesn't blur.

```
if (!_this2.justSelectedSuggestion) {
    _this2.onBlur();
    _this2.onSuggestionsClearRequested();
}
```
It is also linked to https://github.com/moroshko/react-autosuggest/issues/412

At least for mobile, a fix is to do this:

Look for these handlers

```
onMouseEnter: _this3.onSuggestionMouseEnter,
onMouseLeave: _this3.resetHighlightedSuggestionOnMouseLeave,
onMouseDown: _this3.onSuggestionMouseDown,
onTouchStart: _this3.onSuggestionMouseDown, // Because on iOS `onMouseDown` is not triggered
```
      
And add this underneath
`onTouchMove: _this3.onTouchMove, `

Yours might look a little different as I'm just working off the dist code.

Now add this function  below the this.onSuggestionMouseDown function
```
this.onTouchMove = function (e) {
    _this3.justSelectedSuggestion = false;      
  };
```

This means that moving your finger while touching will allow focus again.

However to fully fix this blur() blocking it might need some heavy changes or to catch other cases like this.

scotta-koorong:
Hey all,

The issue is that justSelectedSuggestion is set to true while you scroll and then blocks the .blur() in the onBlur function when you try and blur it afterwards. The displaying of the suggestions is not intrinsically bound to whether the suggestions are focused at the code level. So if justSelectedSuggestion is true then it can't blur.

[...]

PR #417, opened 2017-08-03
Fixes sticky suggestions for touch devices

### LINKED ISSUE #412: Suggestions list stuck open after dragging a suggestion outside of the autosuggest
Issue from here: https://github.com/callemall/material-ui/issues/7526#

d-a-n:
When scrolling on the suggestion list with touch devices and the input looses focus, the suggestions don't disappear as expected.

Fixes  #412 #380


moroshko:
Thanks for this!

Does this really fix #412? 
Isn't `onTouchEnd` a mobile only thing?
Did you test [this repro](https://github.com/moroshko/react-autosuggest/issues/412#issuecomment-318627754) on desktop?

d-a-n:
@moroshko Thanks for you fast reply. I've pushed the missing part ;)

ezgif-4-76db2099a5

moroshko:
@d-a-n I deployed your version [here](https://ra.now.sh/), but it doesn't seem to behave as expected. Specifically:

1. In [#412 repro](https://github.com/moroshko/react-autosuggest/issues/412#issuecomment-318627754), the last bit doesn't work:
> Clicking outside of the autosuggest closes the suggestions. 

2. #380 -Trying to click outside of the autosuggest **on iOS**, doesn't close the suggestions list. It works on desktop, but not on iOS (tested on iPhone 6+ Chrome).

d-a-n:
@moroshko Thanks for pointing this out.

I've fixed `Clicking outside of the autosuggest closes the suggestions.` from #412.

Regarding blurring the the input on touch devices when a users clicks outside: This could be done too but atm it reflects the default browser behavior. I would not change this. Probably most of the users will use the keyboards done or enter key to dismiss the suggestion list anyways. What do you think?

moroshko:
@d-a-n 

#412 
Check out The **Multiple sections** example [here](https://dist-usikpayhrn.now.sh/).
We have `highlightFirstSuggestion={true}` in this example.
The problem is that, with your changes, the first suggestion is highlighted after you drag a suggestion. We should not highlight the first suggestion in this case.

#380
I'm not saying we should blur the input when user clicks outside. I'm saying that the suggestions list needs to be closed. I'm OK with leaving the input focused.
If you still want to fix this, I suggest to create a separate PR for #380.

Issue #332, opened 2017-03-02
Fire onSuggestionSelected when Tab key is pressed


Jackleberry:
I would like to add the ability to fire onSuggestionSelected when the Tab key is pressed. At the moment it only fires on pressing Enter or clicking.

This is the PR: https://github.com/moroshko/react-autosuggest/pull/333

Jackleberry:
I've made a PR with the change I suggested above as well as the right arrow selection.
The onSuggestionSelected will now fire as well when tab or right arrow are pressed.

@moroshko how do you feel about merging in this functionality?

bbarnell:
@moroshko Please merge this into the master and update the UMD build.

rituvesh:
Hi,

Could you please confirm that the selection from tab is is merged ? If yes then please let me know which version I should install.

With regards,
Rituvesh

bbarnell:
@rituvesh The tab key is working on the UMD build for the 9.3.2 version.

rituvesh:
Thanks a lot!

rituvesh:
Hi,
 I can see the tab was implemented https://github.com/moroshko/react-autosuggest/pull/333/files?diff=split
but when I see the new code now only 'enter' and 'click' is used. 'tab' is removed
`onKeyDown: (event, data) => {
        const { keyCode } = event;

        switch (keyCode) {
          case 40: // ArrowDown
          case 38: // ArrowUp
            if (isCollapsed) {
              if (shouldRenderSuggestions(value)) {
                onSuggestionsFetchRequested({
                  value,
                  reason: 'suggestions-revealed'
                });
                this.revealSuggestions();
              }
            } else if (suggestions.length > 0) {
              const {
                newHighlightedSectionIndex,
                newHighlightedItemIndex
              } = data;

              let newValue;

              if (newHighlightedItemIndex === null) {
                // valueBeforeUpDown can be null if, for example, user
                // hovers on the first suggestion and then pressed Up.
                // If that happens, use the original input value.
                newValue =
                  valueBeforeUpDown === null ? value : valueBeforeUpDown;
              } else {
                newValue = this.getSuggestionValueByIndex(
                  newHighlightedSectionIndex,
                  newHighlightedItemIndex
                );
              }

              this.updateHighlightedSuggestion(
                newHighlightedSectionIndex,
                newHighlightedItemIndex,
                value
              );
              this.maybeCallOnChange(
                event,
                newValue,
[...]

PR #427, opened 2017-08-21
fix: Fire onSuggestionSelected when Tab key is pressed

jarlah:
Trying to implement what has been discussed in issue #332 .

Does not build, because i am not good at finding out how to cover that tab key change. Fails due to 90.68 is lower than 91 % test coverage :P

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -582,9 +582,9 @@ export default class Autosuggest extends Component {
 
             break;
 
-          case 9: // Tab
+          // Tab + enter
+          case 9:
           case 13: {
-            // Enter
             // See #388
             if (event.keyCode === 229) {
               break;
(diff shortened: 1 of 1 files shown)


jarlah:
Could potentially refactor stuff so that it becomes:

```
case 9:
case 13:
  this.selectHighlightedItem(event);
  break;
```
or something somethning

jarlah:
no comments on this one? is this relevant? or irrelevant?

davityavryan:
Hey @moroshko,

When you are planing to push this?

There should be just Click aaaand it's merged.

mykter:
PR #333 was a more comprehensive change, but it was rejected. For now I'm just patching it locally to get this behaviour.

Would a PR that had it as a configurable property of the component be of interest?

PR #431, opened 2017-08-31
Fix ie getattribute

lPadier:
IE11 and lower do not support the `getAttribute` method on `SVGUseElement`. As such, when clicking on a <use> tag with a Autosuggest mounted on the page, IE throws "Object doesn't support property or method 'getAttribute'"

### DIFF
--- a/demo/src/components/App/App.js
+++ b/demo/src/components/App/App.js
@@ -6,12 +6,13 @@ import Features from 'Features/Features';
 import Examples from 'Examples/Examples';
 import Footer from 'Footer/Footer';
 
-const App = () =>
+const App = () => (
   <div className={styles.container}>
     <Header />
     <Features />
     <Examples />
     <Footer />
-  </div>;
+  </div>
+);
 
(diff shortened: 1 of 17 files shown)


danez:
I was running into the same issue.

aberezkin:
Hello @lPadier. I'm a new maintainer of this project. 

Can you please make sure that issue still persist? 
And what will be a default behavior on IE in that case?

lPadier:
When a `<use>` tag is clicked in internet explorer, the DOM type is a SVGElementInstance. This type has no method `getAttribute`.

in https://github.com/moroshko/react-autosuggest/blob/23cf0b8a69772e82fdb3ff99949ff29149e4b332/src/Autosuggest.js#L287, getAttribute is called on a node, but this node might be a `SVGElementInstance`.

An error is thrown in `onDocumentMouseDown` in this case

aberezkin:
Yeah, looks fine. Thank you for the contribution!

PR #487, opened 2017-12-11
Update README.md with default behaviour for renderSuggestionContainer

davidfurlong:
Just something I needed to know, feel free to not accept the PR if you don't think this is generally useful


aberezkin:
Looks like it's not relevant since we are moving to headless UI.

PR #496, opened 2018-01-17
added Browsable props for open suggestions with click

uxitten:
this props for open suggestions with user click


aberezkin:
What problem are you trying to solve?

uxitten:
This is about 2 years ago and I don't remember 😀

aberezkin:
Fair notice. I'll close this PR as stale then :)

uxitten:
I think this PR about possibility user to browsing a list without type anything

PR #505, opened 2018-02-08
Add optional selectOnTab prop

Tarpsvo:
Added optional `selectOnTab` property that makes the tab key act identical to the enter key. This feature has been requested multiple times before (as seen in #164, #163, #34, #332 etc). This pull requests adds the named property as well as multiple tests.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -81,7 +81,8 @@ export default class Autosuggest extends Component {
     focusInputOnSuggestionClick: PropTypes.bool,
     highlightFirstSuggestion: PropTypes.bool,
     theme: PropTypes.object,
-    id: PropTypes.string
+    id: PropTypes.string,
+    selectOnTab: PropTypes.bool
   };
 
   static defaultProps = {
@@ -92,7 +93,8 @@ export default class Autosuggest extends Component {
     focusInputOnSuggestionClick: true,
     highlightFirstSuggestion: false,
     theme: defaultTheme,
-    id: '1'
+    id: '1',
+    selectOnTab: false
   };
 
   constructor({ alwaysRenderSuggestions }) {
@@ -581,6 +583,9 @@ export default class Autosuggest extends Component {
 
             break;
 
+          // Tab (must fall through to enter when conditions apply!)
+          case 9:
+            if (!this.props.selectOnTab) break;
           // Enter
           case 13: {
             // See #388
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -586,7 +586,9 @@ export default class Autosuggest extends Component {
           // Tab (must fall through to enter when conditions apply!)
           case 9:
             if (!this.props.selectOnTab) break;
+
           // Enter
+          // eslint-disable-next-line no-fallthrough
           case 13: {
             // See #388
             if (event.keyCode === 229) {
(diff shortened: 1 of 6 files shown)


najisawas:
Hi. Will this be merged soon? I was going to work on it until I noticed this diff :)

src/Autosuggest.js
```diff
@@ -581,7 +583,12 @@ export default class Autosuggest extends Component {
 
             break;
 
+          // Tab (must fall through to enter when conditions apply!)
+          case 9:
+            if (!this.props.selectOnTab) break;
```
najisawas: IMO, I believe Tabbing is heavily used as an "autocomplete" feature. That being said I feel that tabbing should automatically select the first item in the suggestions list if none are highlighted. "Enter", however should continue to remain dependent on a highlighted suggestion

README.md
```diff
@@ -621,6 +622,11 @@ When rendering multiple Autosuggest components on a page, make sure to give them
 <Autosuggest id="destination" ... />
 ```
 
+<a name="selectOnTabProp"></a>
+#### selectOnTab (optional)
+
+Makes the tab key act identical to the enter key, allowing it to select the highlighted suggestion.
```
najisawas: See other comment

dangell7:
Is this still not resolved? Just FYI, outside of the coding world, EVERYONE uses tab before they use ENTER. Most data entry is done very fast and the need for tabbing is far more required than the need for enter.

src/Autosuggest.js
dangell7: I don't. I would suggest that using the up/down arrow key to select then tab selects the item.

Tarpsvo:
Not resolved indeed. I've just been using my own fork in my projects. Works perfectly.

src/Autosuggest.js
DanielaValero: Hello guys, 
I will migrate to this component at the moment, and I've tested the current version with screen readers. The long story made short, the suggestions are not read. I will most likely prepare a PR to improve it in that sense. 

As of the keyboard requirements for accessibility, the UX rules normally are:
Keyboard Interaction

* UP and DOWN: Cycles through auto-suggestions and input field.
* ESC: Close the list with options
* ENTER: Select the currently focused auto-suggestion and close the menu.
* TAB: Select the currently focused auto-suggestion, close the menu, and move focus to the next focusable element.

I hope this helps to figure out the expected behaviour

aberezkin:
Hi! Thank you for the contribution.

You mentioned a couple of old PRs that were closed because it's not quite obvious what should be accessible behavior after pressing `tab` key. As @moroshko [pointed out](https://github.com/moroshko/react-autosuggest/pull/333#issuecomment-308896035) the default behavior in web is to focus next focusable element which is the way it works now in react-autosuggest. Also there are some interfaces that implement behavior equivalent to pressing `down` key. [Chrome search bar for example](https://github.com/moroshko/react-autosuggest/issues/163#issuecomment-217803327). 

So even if we make it optional it's not quite obvious that it should work like `enter`. So what is your motivation to make it behave like `enter`?

Tarpsvo:
It just makes sense to me and was requested by the client and designer in the project I used `react-autosuggest` in. Judging by the popularity of this PR, it makes sense to others as well.

aberezkin:
What if someone’s manager wants it to behave like down key? 

I think we should make the behavior customizable. Not sure how to implement it properly for now. But I think we should look in this direction and research how it’s implemented in other similar libraries.

Issue #503, opened 2018-02-01
event.preventDefault() seems to be ignored


w35l3y:
https://codepen.io/anon/pen/bLVwbM

* Provide the steps to reproduce the issue, e.g.:
  1. Focus on the input field
  2. Type `c`, and wait for suggestions to appear
  3. Click on the suggestion `C++`

  **Observed behaviour:** Suggestion `C++` is selected and suggestions are closed

  **Expected behaviour:** Shouldn't select any suggestion and should keep the suggestions open as the `event.preventDefault()` is called on `onSuggestionSelected` for any suggestion.

moroshko:
What's your use case for not selecting the suggestion when it is clicked?

w35l3y:
I would like to let the user know that some options exist but prevent them from being selected in case the option is present in a secondary list.

devand123:
I did a workaround for this myself. May not be the best way to do it, but figured I'd share:

```
class CustomAutosuggest extends Autosuggest {
  onSuggestionClick = (event) => {
    const {alwaysRenderSuggestions, focusInputOnSuggestionClick} = this.props;
    const {sectionIndex, suggestionIndex} = this.getSuggestionIndices(
      this.findSuggestionElement(event.target)
    );
    const clickedSuggestion = this.getSuggestion(sectionIndex, suggestionIndex);
    const clickedSuggestionValue = this.props.getSuggestionValue(
      clickedSuggestion
    );

    const next = (e) => {
      this.onSuggestionSelected(e, {
        suggestion: clickedSuggestion,
        suggestionValue: clickedSuggestionValue,
        suggestionIndex: suggestionIndex,
        sectionIndex,
        method: 'click'
      });

      if (!alwaysRenderSuggestions) {
        this.closeSuggestions();
      }

      if (focusInputOnSuggestionClick === true) {
        this.input.focus();
      } else {
        this.onBlur();
      }

      setTimeout(() => {
        this.justSelectedSuggestion = false;
      });
    };

    // handle custom onClick
    if (clickedSuggestion && clickedSuggestion.onClick && typeof clickedSuggestion.onClick === 'function') {
      return clickedSuggestion.onClick(next.bind(this, event));
    }

    this.maybeCallOnChange(event, clickedSuggestionValue, 'click');
    next(event);
  }
}
```

[...]

PR #518, opened 2018-02-28
Add onSuggestionClick interceptor function to specify custom behavior.

devand123:
https://github.com/moroshko/react-autosuggest/issues/503

Add the ability to perform custom behavior upon click of a suggestion.

In my case, I wanted to add some search help inline which when clicked upon, it would open a modal. I figure there could be many cases where there is a desire for finer-grained control over what happens when a suggestion is clicked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ class Example extends React.Component {
 | [`getSuggestionValue`](#getSuggestionValueProp) | Function | ✓ | Implement it to teach Autosuggest what should be the input value when suggestion is clicked. |
 | [`renderSuggestion`](#renderSuggestionProp) | Function | ✓ | Use your imagination to define how suggestions are rendered. |
 | [`inputProps`](#inputPropsProp) | Object | ✓ | Pass through arbitrary props to the input. It must contain at least `value` and `onChange`. |
+| [`onSuggestionClick`](#onSuggestionClickProp) | Function | | If specified, will be called whenever a suggestion is clicked. |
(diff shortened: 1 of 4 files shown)


mugiseyebrows:
I was trying to implement input with multiple space-separated items where each item can trigger autosuggestion and needed this function, but then i realized, that intercepting `onClick` is not enough because input value also changed on keyboard 'up' and 'down' events, and then i realized that i can just filter by `method` argument in `onChange()` handler [codepen](https://codepen.io/mugiseyebrows/pen/BVoyeX)

pimterry:
Hi @devand123, thanks for the contribution! It's a good idea, but I have mixed feelings about the details. I know it's been a while, but if you're still interested in this I'd love to talk this through a bit.

The first thing that jumps to mind, is why this is tied to clicking specifically? React-autosuggest tries to be WAI-ARIA compliant, which includes ensuring that all behaviour is keyboard accessible. In general, keyboard selection should do the same thing as mouse selection, and I'm not sure adding an API with special mouse-only functionality fits well with that goal. Is there a reason this can't be integrated into the existing `onSuggestionSelected`?

It looks like the docs at least here are closely based on `onSuggestionSelected` anyway, and there's still some leftover references in there.

I would hypothetically be possible to extend onSuggestionSelected to make selection optional as you're doing here, for example, by allowing users to explicitly return `false` from that method to reject the selection. That would also be a smaller change, and neatly backward compatible. Would that support your use case?

pimterry:
Hi @devand123, did you have any thoughts on my comment above? I know it's been a while, but I'd love to get this resolved if it'd still be useful to you. No worries if not, but I'll close this PR in a few days.

Issue #517, opened 2018-02-28
Preventing default up/down key behavior even when no suggestions are available.


ShibumiGrant:
Default behavior for pressing the <kbd>up</kbd> and <kbd>down</kbd> keys is being `preventDefault`'ed, even in cases where the autosuggest does not need to be using special handling for them. In `input` elements, this manifests as not being able to quickly jump to the beginning and end of an input using these keys, and in `textarea` and `contenteditable` elements it prevents jumping between lines.

Single-line Input Reproduction Steps: (Use the [Basic example](http://codepen.io/moroshko/pen/LGNJMy))
  1. Focus the input field.
  2. Type `asdf`.
  3. Make sure there are no suggestions available.
  4. Press <kbd>up</kbd>.

  **Observed Behavior:** The cursor does not move.

  **Expected Behavior:** The cursor moves to be in front of the first character in the input.

Multi-line Input Reproduction Steps: (Use this [codepen](https://codepen.io/ShibumiGrant/pen/ZrwEQG))
  1. Focus the input field.
  2. Type `asdf`.
  3. Press enter to create a new line.
  4. Type `asdf` again.
  5. Make sure there are no suggestions available.
  6. Press <kbd>up</kbd>.

  **Observed Behavior:** The cursor does not move.

  **Expected Behavior:** The cursor moves to the line above.

moroshko:
Sounds reasonable. Would you like to contribute a PR?

ShibumiGrant:
Gladly! Let me know if any changes should be made.
Any idea when the next release will be?

cozmy:
@moroshko, do you know when you'd have time to merge @ShibumiGrant's changes? I also have the same issue and the proposed changes work.

MUmer01:
Is the any progress in Preventing default up/down key behavior?

PR #522, opened 2018-03-12
Allow default up / down arrow key behavior when no suggestions are available.

ShibumiGrant:
This is to address Issue #517. In lieu of adding coverage around cursor positions, I've added some around expected behavior for using the <kbd>up</kbd> / <kbd>down</kbd> keys when suggestions aren't available.


pimterry:
Thanks for the contribution! This was a great PR, sorry we didn't handle it at the time. As an update, I have good news & bad news. Bad news: I'm not going to merge this now, but good news: this bug has now been fixed independently in #768, so the underlying issue is now resolved anyway :tada:

PR #553, opened 2018-07-05
new prop: closeSuggestionsOnSelect

plysik:
I've added new prop: 'closeSuggestionsOnSelect'.

It allows controling autoclosing suggestions that happens after suggestion selecting.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -1,721 +1,634 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import shallowEqualArrays from 'shallow-equal/arrays';
+import _ from 'underscore';
 import Autowhatever from 'react-autowhatever';
 import { defaultTheme, mapToAutowhateverTheme } from './theme';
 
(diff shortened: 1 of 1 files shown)


noah-potter:
I think a better prop would be something like `keepSuggestionsOnSelect`, that way it defaults to false

plysik:
done

pimterry:
Hi - sorry for the delay here. This looks like a great suggestion, and a really useful feature, but it's quite hard to review because the change reformats a lot of the existing code. On master you can use prettier to reformat the source, by running `npm run prettier -- src`, which should easily get you back to the standard formatting, and make the changes themselves a bit clearer. Unfortunately there's also now conflicts with the current state of master too, so it's not currently mergeable.

If you're still interested in this change, would you be open to looking into that? I know a lot of time has passed here, so no problem if not, just let us know.

pimterry:
Hi @plysik, did you have any thoughts on the above? I'd love to get this merged if it's still useful to you. If not, no worries, but I will close this PR in a few days.

PR #561, opened 2018-08-16
Update react-autowhatever to v10.1.2

KendallWhitman:
### Update react-autowhatever to v10.1.2

This PR patch updates the `react-autowhatever` package for accessibility improvements in OSX mentioned [here](https://github.com/moroshko/react-autowhatever/pull/40) in much greater detail, but mainly moves some aria tags to the container instead of the input, following the ARIA 1.1 specification.
I've also updated the faililng tests that were occuring because of these changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "prop-types": "^15.5.10",
-    "react-autowhatever": "^10.1.0",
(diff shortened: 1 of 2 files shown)


KendallWhitman:
@moroshko No problem, thank you!

PR #566, opened 2018-09-07
Fix suggestions not disappering under some conditions

### LINKED ISSUE #380: inputProps.onBlur not working on mobile

### LINKED ISSUE #412: Suggestions list stuck open after dragging a suggestion outside of the autosuggest

aberezkin:
In this PR I'm trying to solve the problem I described in this [comment](https://github.com/moroshko/react-autosuggest/issues/380#issuecomment-419333641). I decided to propose a separate PR because both #417 and #507 seems to be abandoned and stale. 

First of all, #380 and #412 are caused by the same code and practically identical, so are solved together in this PR.

There is a thing that nor #417 nor #507 don't take in the account. The thing is that despite we should set `this.justSelectedSuggestion` to `false` if the mouse leaves the suggestion we need to set it back to `true` if mouse enters back because `onClick` is fired in that case and the suggestion is selected.

The third important thing is that since we are changing `onSuggestionMouseDown` event, we need to create separate `onSuggestionTouchStart` event, but because most of the browsers are duplicating `touchstart` event with `mousedown` event we need to prevent `onSuggestionMouseDown` from firing. Usually, you would do `e.preventDefault()` but as @gaearon [pointed out](https://github.com/facebook/react/issues/9809#issuecomment-414072263) in facebook/react#9809 this won't work on latest chrome with current version of React (but will work in future when facebook/react#2043 is merged) so for now we can just check if `this.justSelectedSuggestion` is already `true` to not set `this.pressedSuggestion` cause it causes some bugs in mobile browsers.

`npm run build` seems to fail on coverage summary so I also need to write some tests for that (hopefully tests from #417 will suit) but we already can discuss the changes.

Fix #412, fix #380 
Close #417, close #507

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -462,6 +462,7 @@ export default class Autosuggest extends Component {
   onSuggestionTouchMove = () => {
     this.justSelectedSuggestion = false;
     this.pressedSuggestion = null;
+    this.input.focus();
   };
 
   itemProps = ({ sectionIndex, itemIndex }) => {
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -109,10 +109,13 @@ export default class Autosuggest extends Component {
 
     this.justPressedUpDown = false;
     this.justMouseEntered = false;
+
+    this.pressedSuggestion = null;
   }
 
   componentDidMount() {
     document.addEventListener('mousedown', this.onDocumentMouseDown);
+    document.addEventListener('mouseup', this.onDocumentMouseUp);
 
     this.input = this.autowhatever.input;
     this.suggestionsContainer = this.autowhatever.itemsContainer;
@@ -169,6 +172,7 @@ export default class Autosuggest extends Component {
 
   componentWillUnmount() {
     document.removeEventListener('mousedown', this.onDocumentMouseDown);
+    document.removeEventListener('mouseup', this.onDocumentMouseUp);
   }
 
   updateHighlightedSuggestion(sectionIndex, suggestionIndex, prevValue) {
@@ -332,6 +336,9 @@ export default class Autosuggest extends Component {
 
   onSuggestionMouseEnter = (event, { sectionIndex, itemIndex }) => {
     this.updateHighlightedSuggestion(sectionIndex, itemIndex);
+    if (event.target === this.pressedSuggestion) {
+      this.justSelectedSuggestion = true;
+    }
 
     this.justMouseEntered = true;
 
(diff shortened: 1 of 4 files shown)


moroshko:
@aberezkin Thanks so much for taking the initiative to research and fix this long-standing issue!

Changes look good. Let's just make sure that the tests pass, and if you could add a test for this specific bug that would be appreciated!

aberezkin:
@moroshko, I added some tests and now `npm run build` seems to be happy. 

First 3 tests in `when suggestion is dragged` describe section are not passing without changes introduced in this PR and last 2 are needed to prevent degradation.

moroshko:
@aberezkin Thanks again for the effort!
It's published as [v9.4.1](https://github.com/moroshko/react-autosuggest/releases/tag/v9.4.1) now.

aberezkin:
@moroshko Nice, thank you.

src/Autosuggest.js
```diff
@@ -427,18 +446,34 @@ export default class Autosuggest extends Component {
...
+  onSuggestionTouchStart = () => {
+    this.justSelectedSuggestion = true;
+    // todo: event.preventDefault when https://github.com/facebook/react/issues/2043
+    // todo: gets released so onSuggestionMouseDown won't fire in chrome
+  };
+
+  onSuggestionTouchMove = () => {
+    this.justSelectedSuggestion = false;
+    this.pressedSuggestion = null;
+    this.input.focus();
```
tonycassara: Why would you need to refocus the input when `onTouchMove`?

aberezkin: Honestly, I don't remember what exactly happened. But probably touchMove did remove focus from input and that caused suggestions to not disappear on touching something else than input.

tonycassara: Thanks, I ended up extending the library myself to allow suggestions to remain on screen while dismissing keyboard on an iOS/Android device. So far I haven't had any complications due to rewriting this method. Would you be open to a PR changing this?

aberezkin: Of course I will be open to any PR. But I am just a contributor. I think it's to better ask @moroshko about this. Anyway there are tests to prevent degradation in this area. Also, I think it would be reasonable to add a prop to control behaviour introduced by your PR, something like `endureKeyboardDIsmiss={false|true}`.

tonycassara: Yes exactly!! Cool I will throw something together this week.

PR #572, opened 2018-09-13
Fix unwanted input focus after mouse up event on document

### LINKED ISSUE #568: 9.4.1 introduces mousedown/mouseup bug on click to select
## Are you reporting a bug?
Yes.

https://codepen.io/anon/pen/KxQovE

(Note, this required changing the loaded version on Codepen to 9.4.1).

  1. Focus on the autosuggestion input field.
  2. Type `c`, and wait for suggestions to appear
  3. CLICK on a suggestion.
  4. Now click on the second input box below.

  **Observed behavior:** 
Focus goes back to autosuggestion.

  **Expected behavior:**
Focus should not go back to autosuggestion.  This began in 9.4.1 today.

aberezkin:
Autosuggest didn't set `this.pressedSuggestion` to null if `this.justSelectedSuggestion` is `true`. That's is the case when we are just clicking a suggestion in the dropdown. 

I wrote a test to prevent this regression again. My first intention was to put it in `plain-list` tests section but it seems that `blurInput()` helper doesn't take effect if `focusInputOnSuggestionClick` is `true` (`isInputFocused` returns `true`) and test failed even with new changes. So I put it under `do-not-focus-input-on-suggestion-click` section where it fails without change and passes with it.

Fixes #568

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -354,9 +354,9 @@ export default class Autosuggest extends Component {
 
   onDocumentMouseUp = () => {
     if (this.pressedSuggestion && !this.justSelectedSuggestion) {
-      this.pressedSuggestion = null;
       this.input.focus();
     }
+    this.pressedSuggestion = null;
   };
 
   onSuggestionMouseDown = event => {
--- a/test/do-not-focus-input-on-suggestion-click/AutosuggestApp.test.js
+++ b/test/do-not-focus-input-on-suggestion-click/AutosuggestApp.test.js
@@ -6,7 +6,8 @@ import {
   syntheticEventMatcher,
   clickSuggestion,
   focusAndSetInputValue,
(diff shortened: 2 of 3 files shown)


alcalyn:
Thank you, this PR fixes the bug. For now, I'll use 9.4.0 until this merged.

moroshko:
Thanks @aberezkin for fixing this!

Published as [v9.4.2](https://github.com/moroshko/react-autosuggest/releases/tag/v9.4.2) now.

amdrewnguyen:
THANK YOU @aberezkin! Was racking my brain since yesterday trying to figure this one out.

amdrewnguyen:
@moroshko when will this be available on npm?

sedenardi:
@moroshko any chance you could publish this new version to npm?

PR #578, opened 2018-09-21
Don't focus input on mouseup

foodforarabbit:

After clicking on a suggestion, clicking anywhere else triggers onDocumentMouseUp (even after a period of time, I guess the component isn't being unmounted and thus the event listener isn't being removed). `this.justSelectedSuggest` is false / undefined which causes the input to be focussed again. This fix isn't a perfect fix, but it will prevent the input from refocusing if the `focusInputOnSuggestionClick` prop is set to false.

I've built and have this working in my project, and no changes to any tests were required.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -353,8 +353,12 @@ export default class Autosuggest extends Component {
   };
 
   onDocumentMouseUp = () => {
+    const { focusInputOnSuggestionClick } = this.props;
+
     if (this.pressedSuggestion && !this.justSelectedSuggestion) {
-      this.input.focus();
+      if (focusInputOnSuggestionClick) {
+        this.input.focus();
+      }
     }
     this.pressedSuggestion = null;
   };


pimterry:
Thanks for the contribution @foodforarabbit, I'm just digging through old PRs here to clean them up. I'm not sure about this description though:

> After clicking on a suggestion, clicking anywhere else triggers onDocumentMouseUp (even after a period of time, I guess the component isn't being unmounted and thus the event listener isn't being removed)

From looking at the code, it certainly appears that this shouldn't happen - `pressedSuggestion` is set only on mouse down on a suggestion, and is removed on the next mouse up anywhere, so you should never see a fresh click elsewhere triggering the focus change that `pressedSuggestion` causes.

I know it's been a while, but do you have any more detail on the specific issue, or how we can reproduce it? It'd be much better to fix this properly, if it still exists, rather than adding this new quick prop to disable the correct input focus behaviour because of a bug.

pimterry:
Hi @foodforarabbit are you still interested in this? I'd love to get this issue resolved for you, if you're still facing it, and you've got a little time to look into my comment above? No worries if not, but I'll probably close this PR in a few days.

foodforarabbit:
Hmm sorry I haven't worked with this in a long time, not sure how to reproduce anymore. I think it is ok to close this, maybe initial issue was just some version mismatch or some other bug in my code. Thanks!

pimterry:
No problem, thanks for letting us know! :+1:

PR #590, opened 2018-10-31
fix: README.md links and anchor names don't work when in camelCase

kellylu29:


moroshko:
Thanks for fixing this!

PR #602, opened 2019-01-10
Fix event consuming for textareas

RoiEXLab:
I'm using this component for multiline TextFields (i.e. `<textarea>Value</textarea>` in raw html)
And it works, but there is this slight annoyance that every time I press enter to select a suggestion using the keyboard, I add another line to the textarea.
Also when no suggestions are being displayed, I want to be able to freely navigate through the textarea by using arrow keys.
This is the simplest fix i could come up with to prevent that.


RoiEXLab:
Hey @moroshko 
It would be nice if this PR could get merged in the next week or something.
It's a really minor changes, but I'm currently depending on it.
Thanks in advance 👍

RoiEXLab:
Basically a "superset" of #522
See #517

tlginn:
Is there a plan to get this (or https://github.com/moroshko/react-autosuggest/pull/522) merged in the near future? I'm currently using a forked version of the library and would like to be able to update to latest.

hugoholgersson:
Even now, after #768 has landed, I think your fix for Enter makes sense. I think we should land it. But first, it would be great if you could add a test. #768 shows how `preventDefault` can be tested.

pimterry:
Hi @RoiEXLab are you still interested in this? I'd be happy to merge it if you've got a minute to rebase it and add a test. If not no worries, but I'll close this PR in a few days.

RoiEXLab:
Hey @pimterry rebasing the PR is easy, just did it.
Not quite sure about the testing though. I had a quick look at #768 but couldn't really figure out where I had to start, I don't really feel like learning about the test suite for a simple one liner :/
So if someone could give me further guidance on what to do I might have another look at it.

pimterry:
Ok, no worries @RoiEXLab. I've added a couple of quick tests here, which should give us a little coverage of this. Thanks for the contribution! :+1:

RoiEXLab:
Thanks, when can we expect a new npm release?

pimterry:
Now released as v10.0.4 :tada:

Issue #559, opened 2018-08-02
Force open autosuggest on input click


ericbiewener:
## Are you making a feature request?

I have created an `AutoSuggestMulti` component that allows the user to select multiple values. I also have a wrapping `StaticAutoSuggestMulti` component that provides the autosuggest with a fixed set of suggestions, filtered by the query. Therefore, when the query is empty, it provides the full list of suggestions.

Currently, when the user makes a selection, the suggestions become hidden and the `StaticAutoSuggestMulti`'s input remains focused. This is all desirable. At this point, the user can reveal the the suggestions again in a few ways:

1. Blur and re-focus the input
2. Start typing
3. Press the key down 

I'd like the user to be able to simply click the input again to reveal the suggestions. This is how you'd reveal the options in a normal `<select>` field, so clicking is a likely way the user will try to get the suggestions to show up again.

Currently, I've got a hack in place that sets an `onClick` handler on the input which simply calls `setState()` with a copy of the suggestions:
```js
this.setState({ suggestions: this.state.suggestions.map(s => ({ ...s })) });
```

I'd like the API to officially support this behavior, such as via a `revealSuggestionsOnClick` boolean prop.

The-Code-Monkey:
@ericbiewener I am struggling with this issue right now I need it so that when i focus the input field it displays the full list below if there is nothing typed are you able to display the full code of what you did to get this to work struggling. 

I plus one this as a feature for the next version

ericbiewener:
@The-Code-Monkey I think you can accomplish this with the current API. I'm guessing you just need to make [`shouldRenderSuggestions`](https://github.com/moroshko/react-autosuggest#shouldRenderSuggestionsProp) always return `true`.

The-Code-Monkey:
Perfect thank you works as intended

PR #648, opened 2019-06-13
Adds renderSuggestionsOnInputClick feature prop

tinyearth:
Fixes moroshko/react-autosuggest#559

This adds a property to reveal suggestions based on input click events and addresses use case described in [issue 559](https://github.com/moroshko/react-autosuggest/issues/559).

Attached is a slightly modified `demo/src/components/App/components/Examples/components/Basic/Basic.js` for quick reference on usage.

[Basic.js.zip](https://github.com/moroshko/react-autosuggest/files/3287704/Basic.js.zip)

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ class Example extends React.Component {
 | [`onSuggestionSelected`](#on-suggestion-selected-prop) | Function | | Will be called every time suggestion is selected via mouse or keyboard. |
 | [`onSuggestionHighlighted`](#on-suggestion-highlighted-prop) | Function | | Will be called every time the highlighted suggestion changes. |
 | [`shouldRenderSuggestions`](#should-render-suggestions-prop) | Function | | When the input is focused, Autosuggest will consult this function when to render suggestions. Use it, for example, if you want to display suggestions when input value is at least 2 characters long. |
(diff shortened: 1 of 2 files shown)


tinyearth:
@moroshko could you please review and provide feedback? 

It's a pretty simple change; but if for some reason it can't be merged into your repo, I'd like to know sooner than later so I can run with another approach. I'm working on a big project that has a dependency on your library.

krisaoe:
@moroshko would it be possible to merge this feature? I would also like to allow the user to view the suggestions after clicking the input.

aberezkin:
You can use `shouldRenderSuggestions` as pointed out in the issue referenced in the PR.

tinyearth:
It's been over a year, and I'm no longer working on this nor do I have access to the repo with the changes—but `shouldRenderSuggestions` was not sufficient in my use case. 

IIRC there's some nuances to the behavior where that prop was not granular enough. I think there's value in this MR and would suggest reconsidering in detail.

krisaoe:
@aberezkin I agree with @tinyearth 
Our use case is for when the focus returns to the input after selecting a suggestion - if a user clicks the input, the expectation is that the suggestions should render. Using `shouldRenderSuggestions={true}` does not provide the correct behavior.

PR #650, opened 2019-06-17
 Make this package work with tools that use ESM

sdegutis:
In particular, https://www.pika.dev/search?q=react-autosuggest

See https://www.pika.dev/blog/pika-web-a-future-without-webpack/ for more details

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "9.4.3",
   "description": "WAI-ARIA compliant React autosuggest component",
   "main": "dist/index.js",
+  "module": "src/Autosuggest.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/moroshko/react-autosuggest.git"


pimterry:
This would be nice to do, but there's a couple of outstanding issues:

* I'm not clear on the current state of play for ESM. It's not clear if the `module` property in package.json is standard (e.g. neither nodejs's [module packaging docs](https://nodejs.org/api/packages.html) nor npm's [package.json docs](https://docs.npmjs.com/files/package.json) reference it). I've heard rumours of lots of churn around this. Does anybody have an up to date reference on the 'right' way to do this for a modern web module?
* Putting that aside, this approach isn't enough unfortunately. The src files do use raw ESM syntax, absolutely, but they also use lots of non-standard syntax which needs compilation before it can be used, e.g. JSX. Whichever format we use to find the modules, we also need to set up a separate babel pipeline that compiles to modern but standard JS features, while preserving the ESM syntax, and then we can reference that as the module output.
* We'd definitely need some kind of tests for this, that ensure the setup works now, and continues to work in future.

Are you still interested in this issue? For now I'll leave it open, in case you or somebody else is interested in investigating ESM support further, I'm happy to look further at PRs to do so if there's interest.

sdegutis:
I'm not personally interested in it anymore. Feel free to close as needed. Thanks 👍

pimterry:
Ok, thanks for letting us know! I'll close this for now, happy to discuss further if anybody else appears who's interested in this though.

Issue #400, opened 2017-07-07
Documentation needs instructions on how to theme


kylegoetz:
The demos all have arcane CSS rules with no documentation. I see in the dist there's a theme.js, but I can't figure out how to import it and use it. So right now because of probably some copy/paste error from one of the demo CSS files to my own CSS the appearance of the suggestions is not aligned properly. Would be nice to know how I could import  `./node_modules/react-autosuggest/dist/theme.js` but can't find it in the documentation. So I think this counts as a bug I guess.

kylegoetz:
Not sure what to do about this issue, whether to close this and create a new, or just leave a new comment, but here's what I discovered:

Styling applied to the parent container will only be applied to the INPUT but not to the DIV that contains the suggestions. So if you have a parent DIV with text-align: center, it will center the INPUT but not the contents of DIV (I think this is because a block-level element isn't supposed to be moved like that with text-align: center).

So there needs to be some fix for this. If I have time, I'll work on it, but I have a newborn daughter and am the primary caregiver, so it might take a few years :)

nombienombie:
+1

I can't for the life of me understand themes with this plugin

oliviertassinari:
You can find a theme intergration example on the [Material-UI documentation](https://material-ui-1dab0.firebaseapp.com/component-demos/autocomplete).

rishabkdoshi:
+1, need better documentation/examples that use .css instead of .less to style

kimhjona:
PSA: url for Material-UI example is now [here](https://material-ui-1dab0.firebaseapp.com/demos/autocomplete/)

oliviertassinari:
@kimhjona Right, sorry, I have updated the link (we just published a new version based on next.js).

kimhjona:
@oliviertassinari thanks, do you have a link for the source code of that demo? Had it open yesterday, but having a bit of trouble finding it today. Not sure if that url is different as well

kimhjona:
ah, found it. [link](https://github.com/callemall/material-ui/blob/v1-beta/docs/src/pages/component-demos/autocomplete/IntegrationAutosuggest.js) for those looking for the source code

AllModdownload:
Nice
<a href="https://apkmedium.com">apkmedium.com</a>

James-E-Adams:
Hi @moroshko! I'm also struggling to work out how to override default themes. 

Here's what I'm doing at the moment: 
```javascript

[...]

PR #673, opened 2019-08-24
README: Add link to example theme

mgalgs:
This should help with #400.


aberezkin:
I don't think people in #400 wan't an example of how the default theme looks like. Looks like they want instructions on how to make their own theme.

PR #683, opened 2019-10-23
Respect Existing onclick Attributes

mwillbanks:
When defining an onclick attribute within an element inside of the
container or otherwise, you cannot trigger an action since the document
onmousedown event will click-jack the element and cause it never to
fire.

This change detects an existing onclick handler on an element and
instead will fire that elements click handler if it is detected.  The
main use case for this is to have a button with a custom action inside
of the container; the code could go farther to ensure it was within the
container, although, it seems that we should respect other onclick
events.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -283,6 +283,11 @@ export default class Autosuggest extends Component {
       event.target;
 
     while (node !== null && node !== document) {
+      if (node.onclick) {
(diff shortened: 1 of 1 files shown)


SZharkov:
Any idea when this PR will be merged?

aberezkin:
What is your use case? Why would you want to have a button in a suggestion?

PR #709, opened 2020-01-13
Fix focus state from server side rendered input

saltas888:
On 9.4.3 version, `isFocused` state param remains false when the input is focused before the onFocus listener has been initialized. So on SSR before the javascript is ready and before react hydrates server's html, there are two cases of having focused input:
a) When `autoFocus` on inputProps is true.
b) When the user focuses on then input rendered by the server.

-------

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -95,11 +95,11 @@ export default class Autosuggest extends Component {
     id: '1'
   };
 
-  constructor({ alwaysRenderSuggestions }) {
+  constructor({ alwaysRenderSuggestions, inputProps }) {
     super();
 
     this.state = {
-      isFocused: false,
+      isFocused: inputProps.autoFocus,
       isCollapsed: !alwaysRenderSuggestions,
       highlightedSectionIndex: null,
       highlightedSuggestionIndex: null,
@@ -119,6 +119,8 @@ export default class Autosuggest extends Component {
 
     this.input = this.autowhatever.input;
     this.suggestionsContainer = this.autowhatever.itemsContainer;
+
+    this.ensureFocusState();
   }
 
   // eslint-disable-next-line camelcase, react/sort-comp
@@ -169,11 +171,25 @@ export default class Autosuggest extends Component {
         });
       }
     }
+
+    this.ensureFocusState();
   }
 
   componentWillUnmount() {
     document.removeEventListener('mousedown', this.onDocumentMouseDown);
     document.removeEventListener('mouseup', this.onDocumentMouseUp);
+  }
+
+  // This works only when the `renderInputComponent` is not set. Otherwise
+  // input ref is out of react-autowhatever control.
+  ensureFocusState() {
+    if (
+      this.input &&
+      document.activeElement === this.input &&
+      !this.state.isFocused
+    ) {
+      this.setState({ isFocused: true });
+    }
   }
 
   updateHighlightedSuggestion(sectionIndex, suggestionIndex, prevValue) {
@@ -542,7 +558,6 @@ export default class Autosuggest extends Component {
           !this.justClickedOnSuggestionsContainer
         ) {
           const shouldRender = shouldRenderSuggestions(value);
-
           this.setState({
             isFocused: true,
             isCollapsed: !shouldRender
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -558,6 +558,7 @@ export default class Autosuggest extends Component {
           !this.justClickedOnSuggestionsContainer
         ) {
           const shouldRender = shouldRenderSuggestions(value);
+
           this.setState({
             isFocused: true,
             isCollapsed: !shouldRender


hugoholgersson:
This change makes sense to me, but it could need a test...

saltas888:
> This change makes sense to me, but it could need a test...

I added some, I hope it's what you were thinking.

test/with-default-focused-input/AutosuggestApp.js
```diff
@@ -0,0 +1,71 @@
+import React, { Component } from 'react';
+import sinon from 'sinon';
+import Autosuggest from '../../src/Autosuggest';
+import languages from '../plain-list/languages';
+import { escapeRegexCharacters } from '../../demo/src/components/utils/utils.js';
+
+const getMatchingLanguages = value => {
```
hugoholgersson, line 7: I see that this helper is copy/pasted to all tests. If you want to give more love to this repo, one idea could be to factor this out to a perhaps `../plain-list/languages`.

test/with-default-focused-input/AutosuggestApp.js
```diff
@@ -0,0 +1,71 @@
...
+      value,
+      onChange,
+      autoFocus
+    };
+
+    return (
+      <Autosuggest
+        suggestions={suggestions}
+        onSuggestionsFetchRequested={onSuggestionsFetchRequested}
+        onSuggestionSelected={onSuggestionSelected}
```
hugoholgersson, line 64: Unused?

hugoholgersson:
This looks good. I cherry-picked your commits down to master and tried that the tests did pass with your fix and didn't pass without your fix. :) I only left one minor review comment.

Before we land this, how about squashing the commits into one and also update the commit message to not specifically mention  version 9.4.3 (since this is a bug also on master, right)?

@moroshko gave me Collaborator access so I can land this changes once that's done. (If you're interested in helping out more on this git, I think you have good chance to become a Collaborator too, if you ask @moroshko.)

test/with-default-focused-input/AutosuggestApp.js
saltas888: Yes, that makes sense. I can try to do it on another PR, tidying up a bit the tests.

saltas888:
> This looks good. I cherry-picked your commits down to master and tried that the tests did pass with your fix and didn't pass without your fix. :) I only left one minor review comment.
> 
> Before we land this, how about squashing the commits into one and also update the commit message to not specifically mention version 9.4.3 (since this is a bug also on master, right)?
> 
> @moroshko gave me Collaborator access so I can land this changes once that's done. (If you're interested in helping out more on this git, I think you have good chance to become a Collaborator too, if you ask @moroshko.)

Commits squashed 👌.
Changed the commit message 👌.

Of course I'd be happy to help with this, so I'll try to reach @moroshko.

hugoholgersson (approved):
Looks good to me.

test/with-default-focused-input/AutosuggestApp.test.js
```diff
@@ -0,0 +1,28 @@
...
+    expectSuggestions(['C', 'C#', 'C++', 'Clojure']);
+    expect(isInputFocused()).to.equal(true);
+  });
+
+  it('should show suggestions for native focus on input', () => {
+    init(TestUtils.renderIntoDocument(<AutosuggestApp />));
+    focusInputNatively();
+    setInputValue('C');
+    expectSuggestions(['C', 'C#', 'C++', 'Clojure']);
+    expect(isInputFocused()).to.equal(true);
```
hugoholgersson, line 26: This assert should come just after `focusInputNatively()` because that's what it tests?

test/with-default-focused-input/AutosuggestApp.test.js
```diff
@@ -0,0 +1,28 @@
...
+  expectSuggestions
+} from '../helpers';
+import AutosuggestApp from './AutosuggestApp';
+
+describe('Autosuggest with default focused input', () => {
+  it('should show suggestions for autoFocus html attribute', () => {
+    init(TestUtils.renderIntoDocument(<AutosuggestApp autoFocus={true} />));
+    setInputValue('C');
+    expectSuggestions(['C', 'C#', 'C++', 'Clojure']);
+    expect(isInputFocused()).to.equal(true);
```
hugoholgersson, line 18: This assert should come just after `init()` because that's what it tests?

test/with-default-focused-input/AutosuggestApp.js
```diff
@@ -0,0 +1,68 @@
...
+});
+
+export default class AutosuggestApp extends Component {
+  constructor() {
+    super();
+
+    app = this;
+
+    this.state = {
+      value: 'C',
```
hugoholgersson, line 43: If you set this value here, you don't need to call `setInputValue('C')` too in each test, right?

test/with-default-focused-input/AutosuggestApp.js
```diff
@@ -0,0 +1,68 @@
...
+    };
+  }
+
+  render() {
+    const { value, suggestions } = this.state;
+
+    const { autoFocus } = this.props;
+    const inputProps = {
+      value,
+      onChange,
```
hugoholgersson, line 54: Does this test really need to spy on `onChange`?

hugoholgersson (changes requested):

test/with-default-focused-input/AutosuggestApp.js
```diff
@@ -0,0 +1,68 @@
...
+      autoFocus
```
hugoholgersson, line 55: nit: I'd just do `this.props.autofocus` here (to save one line).

src/Autosuggest.js
```diff
@@ -119,6 +119,8 @@ export default class Autosuggest extends Component {
 
     this.input = this.autowhatever.input;
     this.suggestionsContainer = this.autowhatever.itemsContainer;
+
+    this.ensureFocusState();
```
hugoholgersson, line 123: This line doesn't seem to be needed. Your tests passes without.

hugoholgersson (changes requested):

test/with-default-focused-input/AutosuggestApp.test.js
```diff
@@ -0,0 +1,28 @@
...
+  init,
+  setInputValue,
+  isInputFocused,
+  focusInputNatively,
+  expectSuggestions
+} from '../helpers';
+import AutosuggestApp from './AutosuggestApp';
+
+describe('Autosuggest with default focused input', () => {
+  it('should show suggestions for autoFocus html attribute', () => {
```
hugoholgersson, line 14: nit: should show suggestions when first rendered with autofocus

test/with-default-focused-input/AutosuggestApp.test.js
```diff
@@ -0,0 +1,28 @@
...
+    init(TestUtils.renderIntoDocument(<AutosuggestApp autoFocus={true} />));
+    setInputValue('C');
+    expectSuggestions(['C', 'C#', 'C++', 'Clojure']);
+    expect(isInputFocused()).to.equal(true);
+  });
+
+  it('should show suggestions for native focus on input', () => {
```
hugoholgersson, line 21: nit: should show suggestions after natively focused

quyctd:
Hi all,

I'm using nextJS and have a problem with autoFocus. Does this pull request have any updates?

PR #729, opened 2020-03-24
Update build and test pipelines

aberezkin:
Updating build and test libraries to match latest versions.

Some tests aren't passing so need some time to investigate.


aberezkin:
Fixed jsdom to 15.1.1 to fix tests because of this issue https://github.com/jsdom/jsdom/issues/2586. Will leave it like that until we rewrite tests completely.

PR #732, opened 2020-04-06
Update peer deps to 16.3.0

zetoke:
It should be 16.3.0 since the package will not work for older versions since 10.0.0


aberezkin:
Thank you very much! I forgot to do it despite the fact that this was my motivation to release a new version 😅

zetoke:
@aberezkin haha, it happens! 😄

PR #735, opened 2020-04-20
fix: refactor code to use getDerivedStateFromProps

### LINKED ISSUE #624: Warning about `componentWillReceiveProps` in React's Strict Mode
React's [Strict Mode](https://reactjs.org/docs/strict-mode.html) warns about "Unsafe lifecycle methods were found within a strict-mode tree".

> `componentWillReceiveProps`: Please update the following components to use static getDerivedStateFromProps instead.

https://github.com/moroshko/react-autosuggest/blob/d626cef720df0ed5a4c7df4d539836b86bda6afa/src/Autosuggest.js#L124-L143

AdrianInsua:
This pr will fix #624 

In order to do this the code is refactored to use `static getDerivedStateFromProps`. Considering that the method is static i've moved the methods used inside old `willReceiveProps` out of the class to avoid code repeats (because we can't call `this` in static methods).

Also i've moved `justPressedUpDown` `justMouseEntered` `justSelectedSuggestion` into state to acces them in the static method.

Finally i've added a new state variable `prevSuggestions` to emulate `prevProps.suggestions` behaviour.

The demo page works fine, but i had some problems with prettier script on build process
 (but it works ok without it).

I`m not a React developer so please check the code and let me know if something is wrong

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -5,10 +5,55 @@ import Autowhatever from 'react-autowhatever';
 import { defaultTheme, mapToAutowhateverTheme } from './theme';
 
 const alwaysTrue = () => true;
-const defaultShouldRenderSuggestions = value => value.trim().length > 0;
+const defaultShouldRenderSuggestions = (value) => value.trim().length > 0;
 const defaultRenderSuggestionsContainer = ({ containerProps, children }) => (
   <div {...containerProps}>{children}</div>
 );
+const updateHighlightedSuggestionAux = (
+  state,
+  props,
+  sectionIndex,
+  suggestionIndex,
+  prevValue
+) => {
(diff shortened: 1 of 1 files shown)


TaiWilkin:
@AdrianInsua Is this still in-progress?

lregla:
@pimterry @FullstackJack @AdrianInsua 
What steps need to take place at this point in time in order to merge in this PR (other than the obvious merge conflict)?
I am seeing a much more severe StrictMode issue (in line with #624) on React version 17.0.1 with react-autosuggest 10.0.4

```
Warning: Using UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code.
Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.
```

PR #754, opened 2020-07-06
 Fix error in IE11 offsetTop of NULL

pavel-yagodin:
Thanks a lot for contributing to react-autowhatever :beers:

image

### DIFF
--- a/src/Autowhatever.js
+++ b/src/Autowhatever.js
@@ -330,13 +330,12 @@ export default class Autowhatever extends Component {
   };
 
   ensureHighlightedItemIsVisible() {
-    const { highlightedItem } = this;
+    const { highlightedItem, itemsContainer } = this;
 
-    if (!highlightedItem) {
+    if (!highlightedItem || !itemsContainer) {
       return;
     }
 
-    const { itemsContainer } = this;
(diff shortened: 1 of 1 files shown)


pimterry (commented):
Thanks for the contribution. I'm not clear from this PR why this is necessary though, can you explain more?

As far as I can see, `itemsContainer` is set in `storeItemsContainerReference`, which is called (as a ref) in every render. The `ensureHighlightedItemIsVisible` function you've fixed here is only ever called in `componentDidMount` and `componentDidUpdate`, which should both always be called after a successful render, so that should already guarantee that `itemsContainer` is always set.

We could merge this PR, and that would make `ensureHighlightedItemIsVisible` less likely to throw an error, but given the above it seems likely that this error is appearing because of some other issue somewhere. It would be better to fix that root cause instead. It's also not clear how this is related to IE11, since there's no browser-specific behaviour anywhere there...

Can you provide any more detail about the issue, and how we can reproduce the problems you're seeing?

pimterry:
Hi @pavel-yagodin, did you have any thoughts on my comment above? If this is still an issue for you, it'd be great to know more so we can get it sorted. If not, no worries, but I will close this PR in a few days.

Issue #759, opened 2020-08-06
[Bug] highlightedSectionIndex does not reset to null when highlightFirstSuggestion changes from true to false


DoloMike:
**`highlightedSectionIndex` does not reset to null when `highlightFirstSuggestion` changes from `true` to `false`**

[stackblitz example for reproducing the issue](https://stackblitz.com/edit/react-7y2kn1?file=index.js)

### Steps to reproduce the issue (and observed behavior):
  1. Focus on the input field (at this point `highlightFirstSuggestion === false` and nothing is highlighted).
  2. Type `c`, and wait for suggestions to appear (at this point `highlightFirstSuggestion === true` and the first suggestion is highlighted).
  3. Press backspace so the input is empty _(at this point `highlightFirstSuggestion === false` and the first suggestion remains highlighted)._

![captured](https://user-images.githubusercontent.com/5944054/89593483-e7e96680-d814-11ea-98b5-c25193124952.gif)

### Expected behaviour:
  When `highlightFirstSuggestion === false` nothing should be highlighted. `highlightedSectionIndex` should be set back to null.

PR #760, opened 2020-08-07
handle highlightFirstSuggestion state changes

DoloMike:
This is to resolve 

> [Bug] highlightedSectionIndex does not reset to null when highlightFirstSuggestion changes from true to false #759

Code was added inside `Autosuggest.js` -> `UNSAFE_componentWillReceiveProps()` to check if `nextProps.highlightFirstSuggestion` had changed. If `nextProps.highlightFirstSuggestion === false` then `this.resetHighlightedSuggestion()` is called. It ensures that nothing is highlighted once `highlightFirstSuggestion` changes from true to false.

A test was added for it under `./focus-first-autosuggestion`. All other tests remain unchanged and continue to pass.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -123,6 +123,13 @@ export default class Autosuggest extends Component {
 
   // eslint-disable-next-line camelcase, react/sort-comp
   UNSAFE_componentWillReceiveProps(nextProps) {
+    // When highlightFirstSuggestion becomes deactivated, if the first suggestion was
+    // set, we should reset the suggestion back to the unselected default state.
+    const shouldResetHighlighting =
+      this.state.highlightedSuggestionIndex === 0 &&
+      this.props.highlightFirstSuggestion &&
+      !nextProps.highlightFirstSuggestion;
+
     if (shallowEqualArrays(nextProps.suggestions, this.props.suggestions)) {
       if (
         nextProps.highlightFirstSuggestion &&
@@ -131,19 +138,16 @@ export default class Autosuggest extends Component {
         this.justMouseEntered === false
       ) {
         this.highlightFirstSuggestion();
+      } else if (shouldResetHighlighting) {
+        this.resetHighlightedSuggestion();
       }
     } else {
       if (this.willRenderSuggestions(nextProps)) {
         if (this.state.isCollapsed && !this.justSelectedSuggestion) {
           this.revealSuggestions();
-        } else if (!nextProps.highlightFirstSuggestion) {
-          this.resetHighlightedSuggestion();
         }
-        if (
-          this.state.highlightedSuggestionIndex === 0 &&
-          this.props.highlightFirstSuggestion &&
-          !nextProps.highlightFirstSuggestion
-        ) {
+
+        if (shouldResetHighlighting) {
           this.resetHighlightedSuggestion();
         }
       } else {
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -137,6 +137,13 @@ export default class Autosuggest extends Component {
         if (this.state.isCollapsed && !this.justSelectedSuggestion) {
           this.revealSuggestions();
         }
+        if (
+          this.state.highlightedSuggestionIndex === 0 &&
+          this.props.highlightFirstSuggestion &&
+          !nextProps.highlightFirstSuggestion
+        ) {
+          this.resetHighlightedSuggestion();
+        }
       } else {
         this.resetHighlightedSuggestion();
       }
--- a/test/focus-first-suggestion/AutosuggestApp.js
+++ b/test/focus-first-suggestion/AutosuggestApp.js
@@ -13,6 +13,12 @@ const getMatchingLanguages = (value) => {
 
 let app = null;
 
+export const getHighlightFirstSuggestion = sinon.spy(
+  (highlightFirstSuggestion, value) => {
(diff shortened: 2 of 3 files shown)


src/Autosuggest.js
```diff
@@ -136,6 +136,8 @@ export default class Autosuggest extends Component {
       if (this.willRenderSuggestions(nextProps)) {
         if (this.state.isCollapsed && !this.justSelectedSuggestion) {
           this.revealSuggestions();
+        } else if (!nextProps.highlightFirstSuggestion) {
+          this.resetHighlightedSuggestion();
```
pimterry, line 140: If `highlightFirstSuggestion` is always false, this will reset the highlighted suggestion any time any other prop changes, which doesn't seem quite right.

I think it should only reset if all of the below are true:

* The new prop sets highlightFirstSuggestion to false
* The old prop had highlightFirstSuggestion set to true
* The first suggestion is currently highlighted

For example, if the first suggestion is highlighted, I manually select the 3rd item, and then `highlightFirstSuggestion` becomes false, I think it's probably better for the suggestion to stay where it was.

That should still work for your case, but ensures this behaviour is limited just to that specific issue, and doesn't affect normal suggestion selection elsewhere.

pimterry (changes requested):
Thanks for the great contribution! This seems logical, the tests & code look good, and sounds like a sensible improvement to me. It would be useful to be a little more precise about the behaviour, just to limit side effects, but once that's done I think this is good to merge.

test/focus-first-suggestion/AutosuggestApp.test.js
```diff
@@ -135,7 +148,7 @@ describe('Autosuggest with highlightFirstSuggestion={true}', () => {
           suggestionValue: 'Perl',
           suggestionIndex: 0,
           sectionIndex: null,
-          method: 'enter'
+          method: 'enter',
```
hugoholgersson, line 151: I understand the commit hook reformatted these lines... Not your fault... Still, to make the history digging and reviewing easier, I think it's better to split reformatting out to its own commit.

pimterry:
Hi @DoloMike any you still interested in this? I'd be happy to merge it if you've got a minute to handle the comments above, if not no worries, but I'll close this PR in a few days.

DoloMike:
Yeah I have been a little tied up but I'll try to handle this week. Thank
you for the great feedback btw.

src/Autosuggest.js
DoloMike: Thanks for the detailed feedback. I agree and have updated the code now.

test/focus-first-suggestion/AutosuggestApp.test.js
DoloMike: I appreciate the feedback. I agree. I've went ahead and done that with my latest push for what it's worth.

DoloMike:
@pimterry Sorry that I just got around to implementing the changes, but they're here. Please let me know if you would like to discuss further. I'll resolve for now.

src/Autosuggest.js
```diff
@@ -136,6 +136,15 @@ export default class Autosuggest extends Component {
...
+        }
+        if (
+          this.state.highlightedSuggestionIndex === 0 &&
+          this.props.highlightFirstSuggestion &&
+          !nextProps.highlightFirstSuggestion
+        ) {
+          this.resetHighlightedSuggestion();
```
pimterry, line 147: It looks like you've added the discussed change here, but also left the previous code too. Is that intentional? I think the `else if` case above shouldn't be here now, unless I'm missing something.

test/focus-first-suggestion/AutosuggestApp.js
```diff
@@ -4,46 +4,52 @@ import Autosuggest from '../../src/Autosuggest';
...
 };
 
 let app = null;
 
-export const getSuggestionValue = sinon.spy(suggestion => {
+export const getHighlightFirstSuggestion = sinon.spy(
+  (highlightFirstSuggestion, value) => {
+    return Boolean(highlightFirstSuggestion && value);
+  }
+);
```
pimterry, line 20: This doesn't make much sense I'm afraid, I think there's a misunderstanding somewhere.

This code defines a function, which you manually call in the tests, but it doesn't have any side effects, so that doesn't do anything. You then pass this function as the `highlightFirstSuggestion` prop below, which is interpreted as truthy, because it's a function, but it's actually never called there. Maybe `sinon.spy` is creating some confusion: all that does is record every call to this function, so you can check in tests whether the function was called.

If you replace this with just `getSuggestionValue = () => {}` it still does the same thing (and the tests still pass), and similarly if you keep this but swap the `true`/`false` values in the test that passes too.

I think you want to toggle the `highlightFirstSuggestion` prop, so that you actually pass `true` or `false` to `<Autosuggest>` below, and control it from the tests, right? You can do that by:

* Exporting a function here that runs `app.setState({ highlightFirstSuggestion: /* true / false */})`
* Using `this.state.highlightFirstSuggestion` as the `highlightFirstSuggestion` prop in `render()` below
* Updating the test code to call your exported function when you want to switch the values.

Does that make sense? Happy to answer any questions.

pimterry (changes requested):

pimterry (approved):
I've fixed up and finished the last few changes from my comments above myself, plus a bonus fix to match issues that the new tests discovered.

Thanks for the PR @DoloMike! This is all good now imo, so I'm going to merge & release :tada:

PR #763, opened 2020-08-23
(Build): Change prebuild to run prettier:all (instead of prettier without file params)

jafin:
Adjust the prebuild step to run `prettier:all` instead of `prettier`

Running `prettier` command would error due to no file selection.


hugoholgersson:
While `prebuild` indeed gives an error today, I doubt we want to reformat all files? Isn't it more convenient to just reformat the last self-made commit (and the currently staged/modified files)? (When developing, you don't want to change other authors' commits anyway...?)

jafin:
@hugoholgersson I guess my intent was that if you go to the effort of establishing formatting standards, you may as well use them, one less thing to worry about while maintaining code consistency.

pimterry:
Thanks for this PR, good move. Sorry this wasn't finished & merged sooner - unfortunately this issue is now resolved [separately](https://github.com/moroshko/react-autosuggest/blob/93650d203b3250991b2c172c1f0e0ab83030f5a8/package.json#L14) (we've fixed the prettier setup by running it just against the src files directly, rather than against every file in the directory) so I'm going to close this, but the overall issue is now resolved, and I really appreciate the help!

PR #768, opened 2020-10-01
Run SpatNav on up/down (in Android TV's WebView)

hugoholgersson:
Run SpatNav on up/down (in Android TV's WebView)

Background:
Spatial Navigation is Chromium's DPAD-mechanics
that runs in Android WebView on Android TV and
if you run Chrome with --enable-spatial-navigation.
SpatNav runs as Chrome's default action on keydown.

Problem:
Once focus goes to a search box, DPAD_UP/DOWN
in Android TV's WebViews can not move focus
outside the search box. Focus gets trapped.

This happens because react-autosuggest *always*
prevents the default action (Spatial Navigation
in Android TV's WebViews).

Solution:
Do not prevent Spatial Navigation from running
when no suggestions are shown.

Testing done:
1) npm start
2) Quit Chrome.
3) chrome --enable-spatial-navigation
     http://localhost:3000/demo/dist/index.html
4) Ensure that the arrow keys can move between
   all fields and links.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -615,6 +615,7 @@ export default class Autosuggest extends Component {
                   reason: 'suggestions-revealed',
                 });
                 this.revealSuggestions();
+                event.preventDefault(); // We act on the key.
               }
             } else if (suggestions.length > 0) {
               const {
@@ -647,7 +648,7 @@ export default class Autosuggest extends Component {
                 newValue,
                 keyCode === 40 ? 'down' : 'up'
               );
-              event.preventDefault(); // Prevents the cursor from moving.
+              event.preventDefault(); // We act on the key.
             }
 
             this.justPressedUpDown = true;
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -647,9 +647,8 @@ export default class Autosuggest extends Component {
                 newValue,
                 keyCode === 40 ? 'down' : 'up'
               );
+              event.preventDefault(); // Prevents the cursor from moving.
             }
-
-            event.preventDefault(); // Prevents the cursor from moving
 
             this.justPressedUpDown = true;
 
--- a/test/plain-list/AutosuggestApp.test.js
+++ b/test/plain-list/AutosuggestApp.test.js
@@ -89,12 +89,12 @@ describe('Default Autosuggest', () => {
       clearEvents();
     });
 
-    it('should let browser handle dpad_down', () => {
+    it('should let browser handle ArrowDown', () => {
       clickDown();
       expectLetBrowserHandleKeyDown();
     });
 
-    it('should let browser handle dpad_up', () => {
+    it('should let browser handle ArrowUp', () => {
       clickUp();
       expectLetBrowserHandleKeyDown();
     });
@@ -173,13 +173,13 @@ describe('Default Autosuggest', () => {
       expectHighlightedSuggestion(null);
     });
 
-    it('should not let browser handle dpad_down', () => {
+    it('should not let browser handle ArrowDown', () => {
       clearEvents();
       clickDown();
       expectDontLetBrowserHandleKeyDown();
     });
 
-    it('should not let browser handle dpad_up', () => {
+    it('should not let browser handle ArrowUp', () => {
       clearEvents();
       clickUp();
       expectDontLetBrowserHandleKeyDown();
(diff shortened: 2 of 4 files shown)


hugoholgersson:
@aberezkin, could you review this change? It unblocks Android TV.

hugoholgersson:
@saltas888, could you review this change?

Especially, could you apply this change locally and try it out on your web site? Does this break anything for you?

src/Autosuggest.js
```diff
@@ -647,10 +647,9 @@ export default class Autosuggest extends Component {
                 newValue,
                 keyCode === 40 ? 'down' : 'up'
               );
+              event.preventDefault(); // Prevents the cursor from moving.
             }
 
-            event.preventDefault(); // Prevents the cursor from moving
```
pimterry, line 652: I think there's a missed case above here for arrow up/down, when `isCollapsed && shouldRenderSuggestions(value)`. We need a `preventDefault()` in there too: if you press up/down and thereby open the suggestions menu, it shouldn't move the cursor like normal.

The case where preventDefault shouldn't be called is just when `isCollapsed && !shouldRenderSuggestions(value)`, i.e. when autosuggest does nothing with the event.

test/plain-list/AutosuggestApp.test.js
```diff
@@ -81,6 +83,23 @@ describe('Default Autosuggest', () => {
     });
   });
 
+  describe('when input field is focused and empty', () => {
+    beforeEach(() => {
+      focusInput();
+      clearEvents();
+    });
+
+    it('should let browser handle dpad_down', () => {
```
pimterry, line 92: These tests look good, but can we leave out the 'dpad' references in the names? That doesn't make much sense without any context, and they're not really the issue.

The problem (and this fix) is about letting the browser handle arrow key events, whether they come from a keyboard or a dpad or whatever else. There's other issues that have tried to do the same thing for similar reasons elsewhere, unrelated to Android TV, like #522 and #602.

pimterry (commented):
Hi @hugoholgersson! @moroshko gave me collaborator access too, I'm planning to do a run through of the PRs today and help clear things up a bit, thought I'd help here :smile:. This would be a great fix, and would help towards closing out #522 and #602 as well. Just a couple of quick comments first.

src/Autosuggest.js
hugoholgersson: Good catch.

test/plain-list/AutosuggestApp.test.js
```diff
@@ -114,7 +114,9 @@ describe('Default Autosuggest', () => {
     });
 
     it('should hide suggestions when Escape is pressed', () => {
+      clearEvents();
       clickEscape();
+      expectDontLetBrowserHandleKeyDown();
```
hugoholgersson, line 119: drive-by addition: Escape's `preventDefault` was previously untested.

src/Autosuggest.js
hugoholgersson: Done.

pimterry (approved):

pimterry:
Looks great! :+1: :+1: :+1:

hugoholgersson:
Thanks @pimterry!

So let's to step up the package version so we can get this change out there? :) Do you have access to publish a new version to npm? I don't...

I see that https://www.npmjs.com/package/react-autosuggest mentions 10.0.2 as the last version. I don't know how this works because our `package.json` still says `"version": "10.0.1"`. We should probably step it up to 10.0.3?

pimterry:
> So let's to step up the package version so we can get this change out there?

Yep, definitely aiming to do that today, I just want to wrap up #773 first, just need to finish some of the test markups.

> Do you have access to publish a new version to npm?

I do! (I think, although I haven't tested it yet)

> I see that npmjs.com/package/react-autosuggest mentions 10.0.2 as the last version. I don't know how this works because our package.json still says "version": "10.0.1". We should probably step it up to 10.0.3?

Oh boy, that's not great. Doing so digging in https://unpkg.com/browse/react-autosuggest@10.0.2/ to https://unpkg.com/browse/react-autosuggest@10.0.1/ and looking through the commits from back in May, it looks like https://github.com/moroshko/react-autosuggest/commit/9f1ce67ac99b2944ca5c2f088ea4906214bc8567 was published as 10.0.2, but that package.json commit is on a branch, and not on the merged commit on master (which is here https://github.com/moroshko/react-autosuggest/commit/43bd217938fc983a3e6b82d7a91c66f7dd1b2587). It's on the PR branch for #740, but committed & pushed _after_ the branch was already merged, easy mistake to make.

I think it's probably best to leave the github tags as they are, given that (since there really is a correct 10.0.2 tag), but yes we should update straight to 10.0.3 when we release proper.

pimterry:
Now published :+1:

PR #645, opened 2019-06-07
Add aria-label for accessibility

judygab:
When I ran accessibility test on AutoSuggest I received issues that the combobox and input fields don't have descriptions. To fix these issues I added aria-labels which are optional and the users can pass the values of it through props. This would also fix (#485  ), (#283 ),  (#642 ). I'm also creating a pull request for [react-autowhatever](https://github.com/moroshko/react-autowhatever) to pass down the aria-label to the suggestions container.

PR #770, opened 2020-10-06
Add aria-label prop for listbox element

sbeam:
Our accessibility audit revealed that an `aria-label` attribute was required on the search results listing given by the `ul[role=listbox` element. These changes make it possible to pass this down as an optional prop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -633,7 +633,7 @@ The following picture illustrates how `theme` keys correspond to Autosuggest DOM
 ![DOM structure](dom-structure.png)
 
 <a name="aria-label-prop"></a>
-#### theme (optional)
+#### ariaLabel (optional)
 
 Aria-label can be added to props, so that the input field of AutoSuggest is accessible. By default it is set to `search`.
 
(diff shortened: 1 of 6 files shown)


hugoholgersson:
So here you seem to re-use some commits from PR #645... This PR might be easier to review if you did it all in one commit that includes an updated README and also new tests.

Especially, the [README-change](https://github.com/moroshko/react-autosuggest/pull/770/commits/215e5fcb0a785ff99f03daa6c4ae9aa482f85b2a) you cherry-picked from #645 is no longer accurate. Here you change the aria-label of the `<ul>`s, but #645 changed the aria-label of the `<input>` search box... 

I understand we use `inputProps` to pass attributes to the `<input>`... So perhaps we could need a new, similar, top-level prop to pass attributes to the `<ul>`s, say `listboxProps`?

sbeam:
@hugoholgersson you are right on all counts. Thanks for looking. I'll take care of that soon as possible

mlp73:
Looks promising. Any chance to see that PR merged? @sbeam

PR #773, opened 2020-10-13
Add a reason parameter to shouldRenderSuggestions

### LINKED ISSUE #772: Allow finer control in shouldRenderSuggestions
## Are you making a feature request?

* Please describe your use case from user journey point of view, e.g.:

In my application, I have a fixed set of suggestions, where a useful list can be offered before the user has entered any input. When the user focuses an empty input I'd still like this to be hidden as standard, but I'd like to show the suggestions just if they press up or down to explicitly see them.

Once one of those is pressed, I'd like the list of suggestions to appear (and then be dismissable as normal via escape, selectable as normal with enter etc.

I've been trying to use `shouldRenderSuggestions` to do this, but unfortunately it's far too general. E.g. setting it to true all the time shows the suggestions as soon as the field is focused, which isn't what I want.

I want to add one specific extra case where suggestions should be rendered, but I don't want to render them 100% of the time, and there's no way to tell the difference between the two without major monkeypatching of react-autosuggest internals.

* If you have ideas how to extend the Autosuggest API to support your new feature, please share!

Fortunately, I think this is quite easy: pass a reason param to `shouldRenderSuggestions`, just like `onSuggestionsFetchRequested`. That allows use cases like mine, whilst being perfectly backwards compatible with current usage.

Conveniently, in most cases that `shouldRenderSuggestions` is called, it's directly followed by a call to fetch the new suggestions, which already takes a reason.

Here's each of the places its currently called:

1. [`onFocus`](https://github.com/moroshko/react-autosuggest/blob/4c714f557a1f45749ab3f6f1b091cdbf18c60fcc/src/Autosuggest.js#L550), followed by suggestion update with 'input-focused'
2. [`onChange`](https://github.com/moroshko/react-autosuggest/blob/4c714f557a1f45749ab3f6f1b091cdbf18c60fcc/src/Autosuggest.js#L579), with 'input-changed'
3. [`onKeyDown`, arrow up/down](https://github.com/moroshko/react-autosuggest/blob/4c714f557a1f45749ab3f6f1b091cdbf18c60fcc/src/Autosuggest.js#L612), with 'suggestions-revealed'
4. [`onKeyDown`, escape](https://github.com/moroshko/react-autosuggest/blob/4c714f557a1f45749ab3f6f1b091cdbf18c60fcc/src/Autosuggest.js#L717) with 'escape-pressed'
5. [`onBlur`](https://github.com/moroshko/react-autosuggest/blob/4c714f557a1f45749ab3f6f1b091cdbf18c60fcc/src/Autosuggest.js#L443) that no update
6. [`willRenderSuggestions`](https://github.com/moroshko/react-autosuggest/blob/4c714f557a1f45749ab3f6f1b091cdbf18c60fcc/src/Autosuggest.js#L335) with no update

For the first 4, I'd suggest using exactly the same reason as `onSuggestionsFetchRequested`, that seems clear & easy.

'input-blurred' would be easy and consistent for the 5th case.

For the last case, I'd suggest:

* Adding a reason to `willRenderSuggestions` too that gets passed through
* In the two places that that's called:
  * In `render()`, calling it in with 'render' as the reason
  * In `UNSAFE_componentWillReceiveProps()`, calling it with 'suggestions-updated'

Would you be open to a PR to implement this?

pimterry:
This pull request:

* Fixes #772, as described in that issue.
* Fixes `npm run build`, by fixing prettier to run against `src`.
  This wasn't working because `npm run prettier` is missing an argument for which files to prettify. Rather than following #763 and running `prettier:all` (and reformatting a huge amount of code) this just prettifies the src files, which appear to be the previous prettier'd content, and are already very close to correct anyway. If this is merged, #763 could be closed too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,15 +406,25 @@ where:
 
 By default, suggestions are rendered when the input isn't blank. Feel free to override this behaviour.
 
-This function gets the current value of the input, and it should return a boolean.
+This function gets the current value of the input and the reason why the suggestions might be rendered, and it should return a boolean.
 
 For example, to display suggestions only when input value is at least 3 characters long, do:
 
 ```js
-function shouldRenderSuggestions(value) {
+function shouldRenderSuggestions(value, reason) {
   return value.trim().length > 2;
 }
 ```
+
+You can use the second `reason` argument to finely control exactly when the suggestions are rendered. The possible values are closely related to those for `onSuggestionsFetchRequested`, plus a few extra cases:
+
+- `'input-changed'` - user typed something
+- `'input-focused'` - input was focused
+- `'input-blurred'` - input was un-focused
+- `'escape-pressed'` - user pressed <kbd>Escape</kbd> to clear the input (and suggestions are shown for empty input)
+- `'suggestions-revealed'` - user pressed <kbd>Up</kbd> or <kbd>Down</kbd> to reveal suggestions
+- `'suggestions-updated'` - the suggestions were updated
+- `'render'` - the component is re-rendering
 
 When `shouldRenderSuggestions` returns `true`, **suggestions will be rendered only when the input is focused**.
 
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -133,7 +133,7 @@ export default class Autosuggest extends Component {
         this.highlightFirstSuggestion();
       }
     } else {
-      if (this.willRenderSuggestions(nextProps)) {
+      if (this.willRenderSuggestions(nextProps, 'suggestions-updated')) {
         if (this.state.isCollapsed && !this.justSelectedSuggestion) {
           this.revealSuggestions();
         }
@@ -328,11 +328,11 @@ export default class Autosuggest extends Component {
     }
   }
 
-  willRenderSuggestions(props) {
+  willRenderSuggestions(props, reason) {
     const { suggestions, inputProps, shouldRenderSuggestions } = props;
     const { value } = inputProps;
 
-    return suggestions.length > 0 && shouldRenderSuggestions(value);
+    return suggestions.length > 0 && shouldRenderSuggestions(value, reason);
   }
 
   storeAutowhateverRef = (autowhatever) => {
@@ -440,7 +440,7 @@ export default class Autosuggest extends Component {
     const { inputProps, shouldRenderSuggestions } = this.props;
     const { value, onBlur } = inputProps;
     const highlightedSuggestion = this.getHighlightedSuggestion();
-    const shouldRender = shouldRenderSuggestions(value);
+    const shouldRender = shouldRenderSuggestions(value, 'input-blurred');
 
     this.setState({
       isFocused: false,
@@ -535,7 +535,7 @@ export default class Autosuggest extends Component {
       ? alwaysTrue
       : this.props.shouldRenderSuggestions;
     const { value, onFocus, onKeyDown } = inputProps;
-    const willRenderSuggestions = this.willRenderSuggestions(this.props);
+    const willRenderSuggestions = this.willRenderSuggestions(this.props, 'render');
     const isOpen =
       alwaysRenderSuggestions ||
       (isFocused && !isCollapsed && willRenderSuggestions);
@@ -547,7 +547,7 @@ export default class Autosuggest extends Component {
           !this.justSelectedSuggestion &&
           !this.justClickedOnSuggestionsContainer
         ) {
-          const shouldRender = shouldRenderSuggestions(value);
+          const shouldRender = shouldRenderSuggestions(value, 'input-focused');
 
           this.setState({
             isFocused: true,
@@ -576,7 +576,7 @@ export default class Autosuggest extends Component {
       },
       onChange: (event) => {
         const { value } = event.target;
-        const shouldRender = shouldRenderSuggestions(value);
+        const shouldRender = shouldRenderSuggestions(value, 'input-changed');
 
         this.maybeCallOnChange(event, value, 'type');
 
@@ -609,7 +609,7 @@ export default class Autosuggest extends Component {
           case 40: // ArrowDown
           case 38: // ArrowUp
             if (isCollapsed) {
-              if (shouldRenderSuggestions(value)) {
+              if (shouldRenderSuggestions(value, 'suggestions-revealed')) {
                 onSuggestionsFetchRequested({
                   value,
                   reason: 'suggestions-revealed',
@@ -714,7 +714,7 @@ export default class Autosuggest extends Component {
 
                 this.maybeCallOnChange(event, newValue, 'escape');
 
-                if (shouldRenderSuggestions(newValue)) {
+                if (shouldRenderSuggestions(newValue, 'escape-pressed')) {
                   onSuggestionsFetchRequested({
                     value: newValue,
                     reason: 'escape-pressed',
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -10,6 +10,14 @@ const defaultRenderSuggestionsContainer = ({ containerProps, children }) => (
   <div {...containerProps}>{children}</div>
 );
 
+const REASON_SUGGESTIONS_REVEALED = 'suggestions-revealed';
+const REASON_SUGGESTIONS_UPDATED = 'suggestions-updated';
+const REASON_SUGGESTION_SELECTED = 'suggestion-selected';
+const REASON_INPUT_FOCUSED = 'input-focused';
+const REASON_INPUT_CHANGED = 'input-changed';
+const REASON_INPUT_BLURRED = 'input-blurred';
+const REASON_ESCAPE_PRESSED = 'escape-pressed';
+
 export default class Autosuggest extends Component {
   static propTypes = {
     suggestions: PropTypes.array.isRequired,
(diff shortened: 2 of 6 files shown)


pimterry:
Hi @hugoholgersson I didn't realise anybody else was maintaining this repo, and I had previously been planning to just directly merge my own PR here. I'm not sure how involved you want to get, but if you are interested in helping with reviews I'd love to know what you think on this one, so I'm not just unilaterally merging my own code :smiley:. Let me know!

README.md
```diff
@@ -406,16 +406,26 @@ where:
...
 For example, to display suggestions only when input value is at least 3 characters long, do:
 
 ```js
-function shouldRenderSuggestions(value) {
+function shouldRenderSuggestions(value, reason) {
   return value.trim().length > 2;
 }
 ```
 
+You can use the second `reason` argument to finely control exactly when the suggestions are rendered. The possible values are closely related to those for `onSuggestionsFetchRequested`, plus a few extra cases:
```
hugoholgersson, line 419: Do you think a "Reason" enum is overkill? It probably is... But it could be good so we make sure these reasons are in-sync.

pimterry: It's personal taste I guess, but personally I'd prefer keeping them as string values rather than formally making them an enum. I could easily extract each string reason into a constant though, which would help us enforce consistency, what do you think?

We should put fixed list of reason strings into the TypeScript either way too. That'd help for TS obviously, but should also get picked up by pure JS usage in VS code and similar I think, so these will be suggested appropriately. I'll make a PR over there once this one is merged.

test/plain-list/AutosuggestApp.test.js
```diff
@@ -580,7 +580,11 @@ describe('Default Autosuggest', () => {
 
     it('should be called with the right parameters', () => {
       focusAndSetInputValue('e');
-      expect(shouldRenderSuggestions).to.be.calledWithExactly('e');
+      expect(shouldRenderSuggestions.callCount).to.equal(4);
```
hugoholgersson, line 583: For completeness, you might wanna test onBlur and the others too.

README.md
hugoholgersson: > I could easily extract each string reason into a constant though, which would help us enforce consistency, what do you think?

Sounds good, if it's not too much work for you. 

> We should put fixed list of reason strings into the TypeScript either way too. 

Where do we have TypeScript? I though all of this git was plain JS.

test/plain-list/AutosuggestApp.test.js
pimterry: Good idea, done :+1:

README.md
pimterry: > Sounds good, if it's not too much work for you.

Not at all, done :+1:

> Where do we have TypeScript? I though all of this git was plain JS.

Oh yes, this repo is, but there's types available as `@types/react-autosuggest` on npm, from here:

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-autosuggest

All TS users will be using those, and VS Code (among others) will [automatically](https://code.visualstudio.com/docs/nodejs/working-with-javascript#_typings-and-automatic-type-acquisition) find and use those for autocomplete and warnings even in JS.

hugoholgersson: Thanks for the pointer. Interesting. (I'm just using Sublime, without type checking, so far...)

test/plain-list/AutosuggestApp.test.js
```diff
@@ -578,13 +578,25 @@ describe('Default Autosuggest', () => {
...
+
+    it('should be called with the right parameters when revealing/hiding suggestions', () => {
+      focusInput();
+      clickUp();
+      clickEscape();
+      expect(shouldRenderSuggestions.callCount).to.equal(3);
+      expect(shouldRenderSuggestions.getCall(0).args).to.deep.equal(['', 'input-focused']);
+      expect(shouldRenderSuggestions.getCall(1).args).to.deep.equal(['', 'suggestions-revealed']);
+      expect(shouldRenderSuggestions.getCall(2).args).to.deep.equal(['', 'escape-pressed']);
     });
```
hugoholgersson, line 601: Do you want to add a test for the use case you describe in the bug? If I understand your use case correctly, `shouldRenderSuggestions('', 'arrow-pressed')` returning true will force all selections to be shown?

src/Autosuggest.js
```diff
@@ -133,7 +133,7 @@ export default class Autosuggest extends Component {
         this.highlightFirstSuggestion();
       }
     } else {
-      if (this.willRenderSuggestions(nextProps)) {
+      if (this.willRenderSuggestions(nextProps, 'suggestions-updated')) {
```
hugoholgersson: Or rather `props-updated`? Or when is `UNSAFE_componentWillReceiveProps` otherwise called? When suggestions update...?

README.md
```diff
@@ -406,16 +406,26 @@ where:
...
+
+- `'input-changed'` - user typed something
+- `'input-focused'` - input was focused
+- `'input-blurred'` - input was un-focused
+- `'escape-pressed'` - user pressed <kbd>Escape</kbd> to clear the input (and suggestions are shown for empty input)
+- `'suggestions-revealed'` - user pressed <kbd>Up</kbd> or <kbd>Down</kbd> to reveal suggestions
```
hugoholgersson, line 425: Not sure `suggestions-revealed` is the best event name actually... I mean, when `shouldRenderSuggestions` gets this, the suggestions are not yet revealed... Perhaps something like `arrow-pressed` is more honest? (Similar to `escape-pressed`).

test/plain-list/AutosuggestApp.test.js
hugoholgersson: Cool!

src/Autosuggest.js
hugoholgersson: Right, suggestions come as a prop... But yeah, this will also be called when other props are changed...

README.md
pimterry: Yeah, I'm not 100% either way here... It's mostly for consistency, because that's the reason used in fetchRequested for the same event.

I'm not sure `arrow-pressed` would be perfectly honest either, as it isn't called every time an arrow is pressed, only when the suggestions are hidden beforehand, and an arrow is pressed to reveal them.

src/Autosuggest.js
pimterry: The other part of the outer if/else here checks if the suggestions are the same, so that this part of the code actually only runs when the suggestions specifically are updated. It is possible that other props were updated as well, but the suggestions were definitely updated either way, and that's presumably the primary event in this update.

test/plain-list/AutosuggestApp.test.js
```diff
@@ -578,13 +578,25 @@ describe('Default Autosuggest', () => {
...
+    it('should be called with the right parameters during input', () => {
       focusAndSetInputValue('e');
-      expect(shouldRenderSuggestions.callCount).to.equal(4);
+      blurInput();
+      expect(shouldRenderSuggestions.callCount).to.equal(5);
       expect(shouldRenderSuggestions.getCall(0).args).to.deep.equal(['', 'input-focused']);
       expect(shouldRenderSuggestions.getCall(1).args).to.deep.equal(['e', 'input-changed']);
       expect(shouldRenderSuggestions.getCall(2).args).to.deep.equal(['e', 'suggestions-updated']);
       expect(shouldRenderSuggestions.getCall(3).args).to.deep.equal(['e', 'render']);
+      expect(shouldRenderSuggestions.getCall(4).args).to.deep.equal(['e', 'input-blurred']);
```
hugoholgersson, line 589: nit: I'd assert after each interaction instead of asserting everything at once in the end.

test/plain-list/AutosuggestApp.test.js
pimterry: Bit fiddly, but I've found a nice way with `sinon.stub` - now done.

test/plain-list/AutosuggestApp.test.js
pimterry: :+1: done

test/plain-list/AutosuggestApp.test.js
```diff
@@ -592,6 +615,42 @@ describe('Default Autosuggest', () => {
...
+        shouldRenderSuggestions.resetBehavior();
+        shouldRenderSuggestions.callsFake(defaultShouldRenderSuggestionsStub);
+      });
+
+      it('should be able to show suggestions only in certain cases', () => {
+        focusInput();
+        expectSuggestions([]);
+
+        clickUp();
+        expectSuggestions([
```
hugoholgersson, line 636: So here we show all possible items, even though the input field is empty, because of `getMatchingLanguages`'s regexp?

test/plain-list/AutosuggestApp.test.js
hugoholgersson: Nice!

test/plain-list/AutosuggestApp.test.js
pimterry: Yes, exactly, because `new RegExp("^", 'i').test("anything")` is always true, and that's what we effectively end up doing with an empty input (and in reality, my application does something pretty similar too).

hugoholgersson: Ok! Good. Just something one might not expect from that function name. But then, all good :)

hugoholgersson (approved):
Looks good to me.

PR #780, opened 2020-12-07
Add `keepSuggestionsOnSelect` prop

moofspb:
Hi @pimterry!
I found [553](https://github.com/moroshko/react-autosuggest/pull/553) and it seems like staled pull request. @plysik has created useful feature and I think it'll help a lot of people.

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -42,6 +42,7 @@ export default class Autosuggest extends Component {
         );
       }
     },
+    keepSuggestionsOnSelect: PropTypes.bool,
     onSuggestionSelected: PropTypes.func,
     onSuggestionHighlighted: PropTypes.func,
     renderInputComponent: PropTypes.func,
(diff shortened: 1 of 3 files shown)


pimterry:
Hi @moofspb! Thanks for resurrecting this, I agree it's a great feature. Two quick points:

* A couple of tests here would be good, to ensure this behaves correctly. Do you have a minute to look at that? Let me know if you need any help with the test framework at all.

* I think as-is this would work perfectly, but what would you think about going slightly further and making this a callback, not just a fixed value? In some cases (actually including in one of my own projects) it'd be useful to collapse the suggestions only for _some_ suggestion selections, not all of them. E.g. this could be `shouldKeepSuggestionsOnSelect(selectedSuggestion)`? That gives a little extra flexibility. What do you think? Not strictly necessary I know, and I don't want to block this change, but it's an extension I personally would find useful, and it's better to do it now rather than try to support both APIs later.

seropaski:
Hi @pimterry! Thanks for ideas!

I add `shouldKeepSuggestionsOnSelect` method. Does it work like you've expected? If so, we'll add some tests for this feature

pimterry:
Yes, that's perfect! Some quick tests would be great and then I think we're good to go, thanks :+1:

seropaski:
@pimterry done

pimterry (approved):

pimterry:
Great work, thanks again @seropaski!

plysik:
Thank you very much guys. I'm sorry for not responding. I had some hard time. I'm glad that it is going live :)

Issue #669, opened 2019-08-13
Uncaught TypeError: _this4.input.focus is not a function


Hajarath0729:
Hi I am using customized Styled Input component for autosuggest in React functional component. I tried to pass ref using useRef and other ways mentioned in different channels. 

Still getting the same issue. 

const inputRef = useRef(null);

const renderInputComponent = inputProps => {
    return (
      <Fragment>
        <InputWithWarning
         {...inputProps}
         ref={inputRef}
        />
      </Fragment>
    );
  };

<Autosuggest
        suggestions={suggestions}
        onSuggestionsFetchRequested={onSuggestionsFetchRequested}
        onSuggestionsClearRequested={onSuggestionsClearRequested}
        getSuggestionValue={getSuggestionValue}
        renderSuggestion={renderSuggestion}
        shouldRenderSuggestions={shouldRenderSuggestions}
        inputProps={inputProps}
        renderInputComponent={renderInputComponent}
        onSuggestionSelected = {onSuggestionSelected}
      />

Uncaught TypeError: _this4.input.focus is not a function
    at Object.onSuggestionClick [as onClick] (Autosuggest.js:729)
    at Item._this.onClick (Item.js:78)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:270)
    at executeDispatch (react-dom.development.js:561)
    at executeDispatchesInOrder (react-dom.development.js:583)
    at executeDispatchesAndRelease (react-dom.development.js:680)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js:688)

Any suggestion?

DanielGibbsNZ:
Is this the same as #646?

Issue #646, opened 2019-06-07
Uncaught TypeError: Cannot read property 'focus' of undefined


Ksandr-vip:
I am met error when I use `renderInputComponent` API.
```
autosuggest.js:698 Uncaught TypeError: Cannot read property 'focus' of undefined
    at Object.onSuggestionClick [as onClick] (autosuggest.js:698)
    at Item._this.onClick (autosuggest.js:2941)
    at Object.ReactErrorUtils.invokeGuardedCallback (react-dom.js:9017)
    at executeDispatch (react-dom.js:3006)
    at Object.executeDispatchesInOrder (react-dom.js:3029)
    at executeDispatchesAndRelease (react-dom.js:2431)
    at executeDispatchesAndReleaseTopLevel (react-dom.js:2442)
    at Array.forEach (<anonymous>)
    at forEachAccumulated (react-dom.js:15423)
    at Object.processEventQueue (react-dom.js:2645)
```

etairi:
Same here. Would be good if this can be fixed.

mturki:
same error for me

bdmason:
renderInputComponent receives a ref in the component, this needs to be attached to your input component. I'm using MUI and this solves it:

```js
const renderInputComponent = inputProps => {
  const { error, label, ref, ...rest } = inputProps

  return (
    <TextField
      error={!!error}
      label={label}
      fullWidth
      InputProps={{
        inputRef: node => {
          ref(node)
        },
      }}
      {...rest}
    />
  )
}
```

You can do this if you're using a basic HTML input tag:
```html
<input ref={ref} />
```

mouthzipper:
same error for me. Downgrading to 9.4.2 fixed the issue but looking forward to a fix in the new version :)

maniranganathan97:
`import React, { Component } from 'react';
import Radium from 'radium';
import './Person.css'
import styled from 'styled-components';
import Aux from '../../../hoc/Auxillary';
import PropTypes from 'prop-types';
// import Person from '../../../Persons/Person/Person';
import Auxillary from '../../../hoc/Auxillary';

class Person extends Component {
    constructor(props){
        super(props);
        this.inputElementRef = React.createRef;
    }

componentDidMount(){
    this.inputElementRef.current.focus();
}

    render() {
        return (
            // <div className="Person" style={style}> 
            <Auxillary>
[...]

PR #782, opened 2020-12-21
Fix _this.input.focus() is undefined issue on mobile

ZicklePop:
On mobile browsers I've seen this problem in our error logs quite often:

```
TypeError: l.input.focus is not a function
1
File "webpack:///./node_modules/react-autosuggest/dist/Autosuggest.js" line 249 col 19 in Autosuggest
_this.input.focus();
```

It's very similar to what has been mentioned in these issues:
https://github.com/moroshko/react-autosuggest/issues/646
https://github.com/moroshko/react-autosuggest/issues/669

I simply wrapped the focus function so that we make sure it is defined before trying to call the function.

We were able to reproduce the bug consistently by scrolling on the dropdown component in a mobile browser or iOS Simulator.

Hope this helps!

### DIFF
--- a/src/Autosuggest.js
+++ b/src/Autosuggest.js
@@ -509,7 +509,9 @@ export default class Autosuggest extends Component {
   onSuggestionTouchMove = () => {
     this.justSelectedSuggestion = false;
     this.pressedSuggestion = null;
-    this.input.focus();
+    if (this.input.focus) {
+      this.input.focus();
+    }
   };
 
   itemProps = ({ sectionIndex, itemIndex }) => {


eehaightSN (approved):

Stephane-Ag:
I've been trying to replicate this for so long. Thank you.
Didn't even know where this issue was coming from. But indeed needed to scroll from the Autosuggest dropdown. It sends that error even though there's nothing to scroll..

carolineBda:
Hello, can we merge this PR ? I also get the same error

ZicklePop:
@carolineBda at this point I'd suggest migrating to another package that is maintained. The other option is to use [patch-package](https://www.npmjs.com/package/patch-package) method but that will increase tech debt in your project even more.

PR #592, opened 2018-11-14
Add containerProps property for arbitrary props on the container

FabienDeshayes:
This PR adds a new property called containerProps to allow customization of props passed down to `Autowhatever`. It exposes the work done here: https://github.com/moroshko/react-autowhatever/pull/45

Hope the `README` update is enough!

Thanks

PR #785, opened 2021-01-07
Add Arbitrary `containerProps` in `Autosuggest` to allow for overriding in `Autowhatever`

aashishnagpal:
This PR adds a new property called `containerProps` in `Autosuggest` to allow customization of props passed down to `Autowhatever` which already has this exposed.

Credit where it is due - The work was initially done by @FabienDeshayes. I have just added the same things and also implement what was asked in review comments in [#592](https://github.com/moroshko/react-autosuggest/pull/592).

I hope this gets merged and released ASAP as I need this in my project. :beers:

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ class Example extends React.Component {
 | [`getSuggestionValue`](#get-suggestion-value-prop)                     | Function |                        ✓                         | Implement it to teach Autosuggest what should be the input value when suggestion is clicked.                                                                                                          |
 | [`renderSuggestion`](#render-suggestion-prop)                          | Function |                        ✓                         | Use your imagination to define how suggestions are rendered.                                                                                                                                          |
 | [`inputProps`](#input-props-prop)                                      | Object   |                        ✓                         | Pass through arbitrary props to the input. It must contain at least `value` and `onChange`.                                                                                                           |
+| [`containerProps`](#container-props-prop) | Object | | Pass through arbitrary props to the container. Useful if you need to override the default props set by Autowhatever, for example, for accessibility. |
 | [`onSuggestionSelected`](#on-suggestion-selected-prop)                 | Function |                                                  | Will be called every time suggestion is selected via mouse or keyboard.                                                                                                                               |
 | [`onSuggestionHighlighted`](#on-suggestion-highlighted-prop)           | Function |                                                  | Will be called every time the highlighted suggestion changes.                                                                                                                                         |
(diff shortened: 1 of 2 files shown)


aashishnagpal:
@pimterry As promised, this is the new PR. 

Thanks a lot again to take a look at this PR. I'd really really appreciate it if this gets merged and released.

README.md
```diff
@@ -162,6 +162,7 @@ class Example extends React.Component {
...
 | [`getSuggestionValue`](#get-suggestion-value-prop)                     | Function |                        ✓                         | Implement it to teach Autosuggest what should be the input value when suggestion is clicked.                                                                                                          |
 | [`renderSuggestion`](#render-suggestion-prop)                          | Function |                        ✓                         | Use your imagination to define how suggestions are rendered.                                                                                                                                          |
+| [`containerProps`](#container-props-prop) | Object | | Pass through arbitrary props to the container. Useful if you need to override the default props set by Autowhatever, for example, for accessibility. |
```
pimterry, line 165: Optional props, like this, should come after the required props in the table somewhere. This should be _below_ `inputProps`, so that it's easy for people to find all the required props when initially getting started.

README.md
```diff
@@ -316,6 +317,19 @@ function renderSuggestion(suggestion) {
 
 **Important:** `renderSuggestion` must be a pure function (we optimize rendering performance based on this assumption).
 
+<a name="container-props-prop"></a>
+
+#### containerProps
```
pimterry, line 322: Just to make sure we remember: once the above is moved down somewhere, this'll need moving to match too

pimterry (changes requested):
Code looks good, just a small point on the docs. 

If you have some spare time a test might be nice too, but I think this is small enough that that's not strictly required if you don't.

aashishnagpal:
Thanks for the review. I'll get this done in a few minutes. Do you still want a single squashed commit or would you just merge all commits as one?

aashishnagpal:
> Thanks for the review. I'll get this done in a few minutes. Do you still want a single squashed commit or would you just merge all commits as one?

Done the changes, @pimterry 😊

Really appreciate your cooperation and support in this. I am so glad I;ll be able to utilize this new prop in my project.

pimterry (approved):

pimterry:
Released as v10.1.0

NanyThery:
I just wanted to drop a comment because I was experiencing issues using the `containerProps` to override or expand the aria-attributes. Using latest version 1.1.0

I used the prop, and nothing happened on the parent `div` (the one that has the "combobox" role). It turns out that the docs say that you don't need to specify an `id` when using a single instance of this component. However, it also says:

>The only reason id exists, is to set ARIA attributes (they require a unique id).

So, if you add the `id` prop to you component, the ARIA attributes show correctly. This, perhaps, should be clarified in the docs 🤔

Issue #358, opened 2017-04-24
Best way to render differently each sections?


mthuret:
Let's say you have several sections for your autocomplete, but the rendering of each one is completely different. What would be the best way to provide different component depending on the section?

Right now the signature of `renderSuggestion` is `function renderSuggestion(suggestion, { query })`, so it means that your rendering needs to be based on the content of the suggestion to decide how to render. Is there another way to handle this?

moroshko:
I'd just add the section info to each suggestion, and use it in `renderSuggestion`, e.g.:

    const suggestions = [
      {
        text: 'France',
        sectionType: 'country'
      },
      {
        text: 'Watermelon',
        sectionType: 'fruit'
      }
    ];

    renderSuggestion(suggestion) {
      switch (suggestion.sectionType) {
        case 'country': return <Country name={suggestion.text} />;
        case 'fruit': return <Fruit name={suggestion.text} />;
      }
    }

Does this help?

mthuret:
Alright so each suggestions should contain their `sectionType`. 

Thanks!

olance:
I'd vote for reopening this issue, for the following reason: even though it's not too complicated to add `sectionType` to your data, when this data directly comes from an API call result, it forces you to enrich that data if you need to know which section a suggestion belongs to.

For instance, I have to do this:

```js
getSectionSuggestions={section => (hits) => hits.map(h => {
        h.section = section;
        return h;
})}
```

Which is a bit cumbersome, when the signature of `renderSuggestion` could simply be `function renderSuggestion(suggestion, { query, section })`.

Would that be a change you'd consider?

Thanks!

olance:
If I may add, the same issue is found with `renderSuggestionsContainer`: there's no information about which section the container should be rendered for, which makes it hard to customize a specific section.

moroshko:
@olance OK, this makes sense!

Would you like to contribute a PR? (Note it will conflict with https://github.com/moroshko/react-autosuggest/pull/384)

moroshko:
I'm not sure I understand the issue with `renderSuggestionsContainer`. The `children` you get in this function are already rendered using `renderSuggestion`. So, if we add `section` to `renderSuggestion`, wouldn't this be enough?

olance:
@moroshko Yeah, actually I misunderstood the purpose of the function... it might be nice to have a `renderSectionContainer` callback as well when you need to customise specific sections... but that's another story! :)

[...]

PR #789, opened 2021-02-22
`renderSectionContainer`

OliverJAsh:
As originally suggested in https://github.com/moroshko/react-autosuggest/issues/358

For context, I needed this to build the following design:

image


gallantdon:
Can this be merged? This would be extremely useful.

frederik-oestergaard:
Also need this for a very customized autosuggestion popup. Currently what we have to do is very hacky and complex considering the simplicity of the problem. 
All we need to do is essentially render one section as flex row instead of flex column based on the type of section, but without this `renderSectionContainer` there is no simple way of achieving that

PR #839, opened 2022-02-09
fix: Cannot read properties of undefined (reading 'focus')

nfort:
In new version React react-autosuggest has problem with access to Ref, because function component can't work with Ref as prop. 
Rename ref => innerRef


Mohammed-Q96:
@nfort -- any idea when will merge?

nfort:
@Mohammed-Q96 no, I can guess this package don't have mantain future, because author not support after 2021. I forked and publish fixed version as @nfort/react-autosuggest

